PR #10, opened 2019-09-28
Fix option (yet to include more tests) and remove list

afsalthaj:


src/main/scala/zio/config/package.scala
```diff
@@ -5,37 +5,24 @@ import java.net.URI
...
-              .fold[Either[WriteError, A]](Left(WriteError.apply("cannot obtain the value", None)))(t => Right(t))
-        )
-    }
+  def opt[A](config: => Config[A]): Config[Option[A]] =
+    config
+      .mapEither[Option[A]](a => Right(Some(a)))({
+        case Some(value) => Right(value)
+        case None        => Left(WriteError("Error: Cannot write a none value", None))
+      })
+      .onError(_ => None)
```
afsalthaj: Create a `mapMaybe`, instead of this trip

src/main/scala/zio/config/package.scala
```diff
@@ -5,37 +5,24 @@ import java.net.URI
...
-    config match {
-      case Config.Sources(propertyType, paths) => Config.Sources(PropertyType.ofOption(propertyType), paths)
-      case Config.Source(path, propertyType)   => Config.Source(path, PropertyType.ofOption(propertyType))
-      case a =>
-        a.errorMap(a => Right(Some(a): Option[A]))(
-          value =>
-            value.fold[Either[WriteError, A]](Left(WriteError.apply("cannot obtain the value", None)))(t => Right(t))
-        )
-    }
+  type ConfigOption[A] = Config[Option[A]]
```
afsalthaj: Remove this

src/test/scala/zio/config/ListAndOptionalTests.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  val cId: Config[Id] = string("kId").xmap(Id)(_.value)
+
+  val cOverallConfig: Config[OverallConfig] =
+    opt(cId).xmap(OverallConfig)(_.list)
+
+  property("optional write") = forAllZIO(genOverallConfig) { p =>
+    write(cOverallConfig).run
+      .provide(p)
+      .shouldBe(KeyValue(p.list.map(t => Map("kId" -> t.value)).getOrElse(Map.empty[String, String])))
```
afsalthaj: `p.map` and not `p.list.map`

README.md
```diff
@@ -19,14 +22,14 @@ Configuration parsing should be easy as it sounds - Hence;
 
 import zio.{App, ZIO}
 import zio.config._
+import zio.console.Console
 
 object ReadConfig extends App {
-  case class Prod(ldap: String, dburl: Option[String], regions: List[Int])
```
leigh-perry: List support to be added later

src/main/scala/zio/config/Config.scala
```diff
@@ -1,36 +1,34 @@
 package zio.config
 
-abstract class Config[A] { self =>
```
leigh-perry: This could (should?) be a sealed trait

src/main/scala/zio/config/actions/Read.scala
```diff
@@ -15,7 +14,7 @@ case class Read[A] private (value: ZIO[(Ref[ConfigReport], ConfigSource), List[R
 
 object Read {
   // Read
-  final def read[A](config: Config[A]): Read[A] =
```
leigh-perry: case FlatMap is not handled

src/main/scala/zio/config/actions/Write.scala
```diff
@@ -6,33 +6,35 @@ import zio.ZIO
 final case class Write[A](run: ZIO[A, WriteError, KeyValue])
 
 object Write {
-
-  // Write
   final def write[A](config: Config[A]): Write[A] =
     config match {
```
leigh-perry: case FlatMap is not handled

src/main/scala/zio/config/Config.scala
afsalthaj: 👍

leigh-perry:
See review comments inline.

src/main/scala/zio/config/actions/Read.scala
afsalthaj: I removed it.
In almost all config libs, flatMap is something inherently wrong or monadically not lawful.
For us, we can add it as a separate feature 
It's easy to build a flatMap meanwhile, but after some discussions.

src/main/scala/zio/config/actions/Write.scala
afsalthaj: FlatMap was removed

README.md
afsalthaj: Yes, list is a prioritised feature. Just a matter, how do user need to express it.

`opt(config)` makes sense intuitively. Optionally read all the values associated with the `config`. 

but what exactly is `list(config)`. Immediate thinking is reading all the values within that config as a list?
So library will assume the following. It will be a delimited string. But is that the only representation? 

`string("possibly-a-list").mapEither(t => t.split(",").toList)(_.mkString(","))` is the implementation. But is that a function in the library?

afsalthaj: In other words,
`list(int("x"))` could possibly mean the following

1. Read the integer first, and then convert it to a list.
2. Or (as library used to do), read the string first, delimit split, convert each of them to an integer.

PR #14, opened 2019-09-29
More property tests

leigh-perry:
@afsalthaj see the TODOs for failing tests - they are commented out for the moment


src/test/scala/zio/config/ReadWriteRoundtripTest.scala
```diff
@@ -72,10 +83,23 @@ object ReadWriteRoundtripTest extends Properties("Coproduct support") with TestS
     p2.shouldBe(p)
   }
 
+  // TODO inhibited failing test until fixed
+  //  property("coproduct roundtrip") = forAllZIO(genCoproductConfig) { p =>
```
afsalthaj: Is it coproduct round trip,,or coproduction with nested ?

src/test/scala/zio/config/OperationsTest.scala
```diff
@@ -0,0 +1,32 @@
...
+  //      .provide(p)
+  //      .shouldBe(KeyValue(Map("kId" -> p.opt.map(_.value).mkString(","))))
+  //  }
+
+  ////
+
+  final case class Id(value: String) extends AnyVal
+  final case class OverallConfig(opt: Option[Id])
+
+}
```
afsalthaj: Isn't it the same as ?
https://github.com/zio/zio-config/blob/master/src/test/scala/zio/config/ListAndOptionalTests.scala

afsalthaj: Also, isn't the expected value wrong?
`.shouldBe(KeyValue(Map("kId" -> p.opt.map(_.value).mkString(","))))`

Take a look at the above test. `kId` won't exist ever, if `p.opt` don't exist.

src/test/scala/zio/config/ReadWriteRoundtripTest.scala
leigh-perry: It is just a coproduct of some level of complexity (the nested stuff)

src/test/scala/zio/config/ReadWriteRoundtripTest.scala
```diff
@@ -72,10 +83,23 @@ object ReadWriteRoundtripTest extends Properties("Coproduct support") with TestS
...
+
   ////
 
-  final case class Ldap(value: String)  extends AnyVal
+  final case class Id(value: String)    extends AnyVal
   final case class DbUrl(value: String) extends AnyVal
-  final case class EnterpriseAuth(ldap: Ldap, dburl: DbUrl)
+  final case class EnterpriseAuth(id: Id, dburl: DbUrl)
   final case class NestedStuff(enterpriseAuth: EnterpriseAuth, count: Int, factor: Double)
+  final case class CoproductConfig(coproduct: Either[Option[Id], NestedStuff])
```
afsalthaj: When you get time, lets renamed to NestedConfig or something along those lines.
I would also not mind creating the config for every test in the property instead of global declarations.
They are only tiny descriptions and recomputing is fine.

src/test/scala/zio/config/ReadWriteRoundtripTest.scala
```diff
@@ -62,20 +67,33 @@ object ReadWriteRoundtripTest extends Properties("Coproduct support") with TestS
...
   ////
 
-  final case class Ldap(value: String)  extends AnyVal
+  final case class Id(value: String)    extends AnyVal
   final case class DbUrl(value: String) extends AnyVal
-  final case class EnterpriseAuth(ldap: Ldap, dburl: DbUrl)
-  final case class NestedStuff(enterpriseAuth: EnterpriseAuth, count: Int, factor: Double)
+  final case class EnterpriseAuth(id: Id, dburl: DbUrl)
+  final case class NestedConfig(enterpriseAuth: EnterpriseAuth, count: Int, factor: Double)
+  final case class CoproductConfig(coproduct: Either[Option[Id], NestedConfig])
```
afsalthaj: When you say `Either[Option[A], B]`, what does it mean.
If A == None, take B?
Option[A] is valid if its None isn't it?
you mean Either[A, B]?

afsalthaj: Also 

```
  val cNestedStuff: Config[NestedStuff] =
    (int("kCount") <*> double("kFactor"))(
      NestedStuff.apply,
      NestedStuff.unapply
    )
```
was forward referenced when defining
```
  val cCoproductConfig: Config[CoproductConfig] =
    (cId2 or opt(cNestedStuff))
      .xmap(CoproductConfig)(_.coproduct)
```

This resulted in null.

And to make much more sense, could you change it to 
`final case class CoproductConfig(coproduct: Either[Id, Option[NestedConfig]])`
That will be a valid round trip.
The previous round trip will always give only `Left(None)` which makes sense to me.

afsalthaj: Either[A, B] makes sense to me and not Either[Option[A], B].,
Either[Option[A], B] is a Left for me always, and it makes sense
Either[A, Option[B]] would be fine. It won't return config error at all

afsalthaj: Although the main error was due to forward reference to null (which I tested and it ran fine)
```
sbt:zio-config> testOnly zio.config.ReadWriteRoundtripTest
[info] + Coproduct support.nested case class roundtrip: OK, passed 100 tests.
[info] + Coproduct support.newtype 1 roundtrip: OK, passed 100 tests.
[info] + Coproduct support.newtype 2 roundtrip: OK, passed 100 tests.
[info] + Coproduct support.case class 1 roundtrip: OK, passed 100 tests.
[info] + Coproduct support.coproduct roundtrip: OK, passed 100 tests.
```

afsalthaj:
@leigh-perry 
I think, we could uncomment all the tests, and make it run instead of merging commented tests.
based on the inputs In the PR comments.

Issue #21, opened 2019-09-29
Change UIO[Option[String]] to IO[Unit, String] in ConfigSource.scala


afsalthaj:
In ConfigSource.scala, change UIO[Option[String]] to IO[Unit, String].
Also, rename `getString` to `getConfigValue`.

wi101:
fixed in #27

PR #27, opened 2019-09-29
Change UIO[Option[String]] to IO[Unit, String] in ConfigSource.scala

wi101:
#21

Moved `<*>`to the trait `Config` and remove `ConfigSyntax`
Changed UIO[Option[String]] to IO[Unit, String] in ConfigSource.scala
Remove Ref from the Environment of the Read value


src/main/scala/zio/config/actions/Read.scala
```diff
@@ -74,25 +58,25 @@ object Read {
...
-        )
 
-      case Config.Xmap(c, from, _) =>
-        Read(read(c).value.map(t => (t._1, from(t._2))))
-    }
+        case Config.Xmap(c, from, _) =>
+          loop(c, report).map(t => (t._1, from(t._2)))
+      }
+
+    Read(Ref.make(ConfigReport(Nil)).flatMap(report => loop(configuration, report)))
```
afsalthaj: Makes perfect sense!

PR #35, opened 2019-09-30
Issue 20 - use List cons as a budget NonEmptyList

leigh-perry:


zio-config/src/main/scala/zio/config/package.scala
```diff
@@ -27,4 +27,19 @@ package object config extends Sources {
...
+  type NonEmptyList[A] = ::[A]
+
+  object NonEmptyList {
+    def apply[A](a: A, tail: A*): NonEmptyList[A] =
+      ::(a, tail.toList)
+
+    def concat[A](lerr: NonEmptyList[A], rerr: NonEmptyList[A]): NonEmptyList[A] =
+      ::(lerr.head, lerr.tail ++ rerr)
+  }
+
```
leigh-perry: Idea here is to have NEL reusable in the future.

afsalthaj: Let's be specific about ReadErrors,

PR #37, opened 2019-09-30
Implement Config.Optional

afsalthaj:


afsalthaj:
@wi101 @jdegoes 
I have included two changes in the PR.
One is including `Config.Optional` and next is changing from `<*>` to `|@|`.
Will ensure a PR address only 1 issue if thats a requirement

zio-config/src/main/scala/zio/config/actions/Write.scala
```diff
@@ -13,8 +13,16 @@ object Write {
...
-      case Config.Xmap(c, _, to) =>
-        Write(ZIO.accessM(b => write(c).run.provide(to(b))))
+      case Config.Optional(c) =>
+        Write(
+          ZIO.accessM(
+            a =>
+              a.asInstanceOf[Option[A]]
+                .fold[ZIO[A, WriteError, Map[String, String]]](
+                  ZIO.succeed(Map.empty[String, String])
+                )(aa => write(c).run.provide(aa))
```
afsalthaj: @jdegoes The place where I had to use `asInstanceOf`.

wi101: didn't understand, the a is of type A right?
I would do this: (but not sure if it would be correct)
```scala
case c @ Config.Optional(_) =>
  write(c).run.fromOption.foldM(_ => UIO(Map.empty[String, String]), UIO(_))
```

Consider to add test

afsalthaj: The optional was already implemented before using `opt` function in package object.
Hence the tests were there before.

afsalthaj: I tried 
```
case c @ Config.Optional(_) =>
  write(c).run.fromOption.foldM(_ => UIO(Map.empty[String, String]), UIO(_))
```
It's actually interesting. But unfortunately the types are still not aligned. `A` comes in the contra variant position for write. And we say, the `A` was actually `Option[A]` during a `Config.Optional` program.

It's almost a natural transformation that we are doing, and with Scala, we are unaware of the inner types in GADT, and we have to explicitly do the cast.

afsalthaj:
@leigh-perry @jdegoes @wi101  Happy to merge now since the PR getting piled up?

PR #42, opened 2019-10-01
Move methods from config companion object to object Config

### LINKED ISSUE #28: Move the methods from config package to the companion object Config
Move: `int`, `double`, `string`,`long`, `short`,`uri`, `read`, `write`, and `report` to the companion object `Config` and make sure the examples still work.

pgabara:
Fixes #28

### DIFF
--- a/zio-config/src/main/scala/zio/config/Config.scala
+++ b/zio-config/src/main/scala/zio/config/Config.scala
@@ -1,5 +1,7 @@
 package zio.config
 
+import java.net.URI
+
 sealed trait Config[A] {
   self =>
   final def zip[B](that: => Config[B]): Config[(A, B)] = Config.Zip(self, that)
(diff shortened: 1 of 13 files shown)


zio-config/src/main/scala/zio/config/Config.scala
```diff
@@ -37,4 +41,21 @@ object Config {
...
+
+  def short(path: String): Config[Short] = Config.Source(path, PropertyType.ShortType)
+
+  def uri(path: String): Config[URI] = Config.Source(path, PropertyType.UriType)
+
+  def read[A](config: => Config[A]): Read[A] = Read.read[A](config)
+
+  def write[A](config: => Config[A]): Write[A] = Write.write[A](config)
+
+  def report[A](config: => Config[A]): Report[A] = Report.report[A](config)
```
afsalthaj: I would imagine the `read`, `write` and `report` to be still part of the package object and rest of the functions here.

The main import for the user could be , for example,
`import zio.config._, Config._`

zio-config/src/main/scala/zio/config/package.scala
```diff
@@ -20,10 +11,6 @@ package object config extends Sources {
       })
       .onError(_ => None)
 
-  def read[A](config: => Config[A]): Read[A]     = Read.read[A](config)
-  def write[A](config: => Config[A]): Write[A]   = Write.write[A](config)
-  def report[A](config: => Config[A]): Report[A] = Report.report[A](config)
```
afsalthaj: Based on the discussions, the read write and report are back in here. We can inline the functions later on though.
@wi101?

zio-config/src/main/scala/zio/config/Config.scala
pgabara: sure, `read`, `write` and `report` moved back.

zio-config/src/main/scala/zio/config/package.scala
wi101: Alright

afsalthaj:
@pgabara If you could resolve the conflicts, we can merge this one ?

pgabara:
I resolved conflicts and added new functions: `byte`, `bigInt`, `boolean` etc to companion object.

PR #47, opened 2019-10-01
Move Read, Write, Report to the package config and make them methods

### LINKED ISSUE #25: Move Read, Write, Report to the package config and make them methods
currently in the `config` package those methods are using different data type

```scala
  def read[A](config: => Config[A]): Read[A]     = Read.read[A](config)
  def write[A](config: => Config[A]): Write[A]   = Write.write[A](config)
  def report[A](config: => Config[A]): Report[A] = Report.report[A](config)
```

Move the implementation of `Read`, `Write` and `Report` to `config` and make them methods instead of data types.
those methods should return `ZIO` 

```scala
  def read[A](config: => Config[A]): ZIO[ConfigSource, List[ReadError], (ConfigReport, A)]  
  def write[A](config: => Config[A]): Map[String, String]
  def report[A](config: => Config[A]): ConfigReport
```

pgabara:
Fixes #25


afsalthaj:
@pgabara @wi101 I would personally prefer waiting for other PRs to get merged before this one goes in. Moving around the lines of code can be a bit of tricky with respect to change as we are in the phase of adding more functionalities into Read, Write, Report.

examples/src/main/scala/zio/config/examples/CoproductExample.scala
```diff
@@ -45,7 +45,7 @@ object CoproductExample extends App {
   val anotherSource: ConfigSource =
     mapSource(validConfigForAnotherConfig)
 
-  assert(runtime.unsafeRun(read(prodOrDev).run.provide(anotherSource))._2 == Right(Dev("v3", 1, 2.0)))
+  assert(runtime.unsafeRun(read(prodOrDev).provide(anotherSource))._2 == Right(Dev("v3", 1, 2.0)))
```
afsalthaj: 👍

afsalthaj: Let's revert the moving of lines of code and deletion of files.
We can do it later. Let this use case be solving one issue of having the return types to be just ZIO

afsalthaj:
We will end up having to change a few things (we may not expose `read` itself and might end up having it as an internal function.
Sorry that, let's close it for now.

pgabara:
I totally understand. Thanks for comments.

PR #44, opened 2019-10-01
Feature/add sequence

afsalthaj:

PR #48, opened 2019-10-01
Added additional types to PropertyType

Michaelt293:
I have added support for additional numeric types as well as boolean values. There are a few things that should be considered-
* do we wish to use the `override` keyword (as has been used for `StringType`)?
* do we wish to have case insensitive boolean values? Currently, the `toBoolean` method is used which is case insensitive.

```
scala> "TRue".toBoolean
res0: Boolean = true
```

I believe we could also consider adding `Option` and `List` to the ADT-
```
  case class OptionType[A](propertyType: PropertyType[A]) extends PropertyType[Option[A]] {
    def description: String = s"optional ${propertyType.description}"
    def read(value: String): Either[ErrorType, Option[A]] =
      if (value.isEmpty) Right(None)
      else propertyType.read(value).map(Some(_))
    def write(value: Option[A]): String = value.fold("")(propertyType.write)
  }
```

### DIFF
--- a/zio-config/src/main/scala/zio/config/PropertyType.scala
+++ b/zio-config/src/main/scala/zio/config/PropertyType.scala
@@ -23,25 +23,18 @@ object PropertyType extends AttemptSyntax {
     override def write(a: String): String                       = a
   }
 
-  case object IntType extends PropertyType[Int] {
-    def description: String = "value of type int"
-    def read(value: String): Either[ErrorType, Int] =
-      value.toInt.attempt(_ => ParseError(value, "int"))
-    def write(value: Int): String = value.toString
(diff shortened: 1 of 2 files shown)


afsalthaj:
HI @Michaelt293 Option and list are considered as just effects, that we compose along with config descriptions.

If you could look at my PR (https://github.com/zio/zio-config/pull/44/files), which implement sequence, we do handle list there. In fact, List was a type in `PropertyType` before and I removed it.

https://github.com/zio/zio-config/pull/44/files#r329900614

Michaelt293:
Awesome, thanks for the information. I knew that Option and List are supported but good to know the design decisions :)

PR #49, opened 2019-10-02
Replace WriteError with String and change ReadErrors.scala

### LINKED ISSUE #38: Replace WriteError with String and change ReadErrors.scala
Rename `Errors.scala` file to `ReadErrors.scala` file;
Remove WriteError. Replace with String in places where WriteError is used.

We will change the encoding of ReadError to the following

```scala
sealed trait ReadError {
  def key: String
}

object ReadError {
 final case class MissingValue(key: String) extends ReadError(key)
 final case class ParseError(key: String, provided: String, `type`: String) extends ReadError(key)
}

```

This will make it consistent with zio

pgabara:
Fixes #38


pgabara:
Thanks!

Issue #57, opened 2019-10-03
Add Empty Config


wi101:
Remove `Pure` Config and add `Empty` Config which describes an empty configuration and preserve the type of the config

```scala
final case class Empty[A]() extends Config
```

afsalthaj:
case class Empty[A](), interpreting it during Read can be a bit tricky.
Let's discuss.

afsalthaj:
#58

afsalthaj:
@wi101  i think we merged this. Hence closing this :)

PR #58, opened 2019-10-03
Issue 57: Added Empty config and removed Pure

wi101:
#57


zio-config/src/main/scala/zio/config/actions/Read.scala
```diff
@@ -15,7 +15,7 @@ object Read {
       report: Ref[ConfigReport]
     ): ZIO[ConfigSource, ReadErrors, (ConfigReport, B)] =
       configuration match {
-        case Config.Pure(value) => report.get.map(t => (t, value))
+        case Config.Empty() => ZIO.fail(ReadErrors(MissingValue("undefined")))
```
afsalthaj: Empty silently pass the previous computation to next? And not fail?

afsalthaj: I think empty makes more sense to me than the name  Pure. However we don’t need to fail the read

wi101: do you mean empty should wrap a value: A ? (like Pure)

afsalthaj: We could have just 
Config.empty[A] which returns a Config[Option[A]]

During read we can feed in just none. 

And in sequence,  instead of Config.Pure(Nil), we could do Config.empty[A].xmapEither(_.toList)(_.headOption)

I am not sure. Just an idea. It will make sense to user when Config.empty[A] is populating Config with a none value. What do you think ?

afsalthaj: I use to depend on monoid during these situations. That we can fold over an F, to form a monoidal A.

wi101: Good idea! Thanks!

zio-config/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -57,11 +57,10 @@ object ConfigDescriptor {
   final case class Or[A, B](left: ConfigDescriptor[A], right: ConfigDescriptor[B])
       extends ConfigDescriptor[Either[A, B]]
 
-  def succeed[A](a: A): ConfigDescriptor[A] =
-    ConfigDescriptor.Succeed(a)
+  def empty[A]: ConfigDescriptor[Option[A]] = ConfigDescriptor.Empty()
 
   def sequence[A](configList: List[ConfigDescriptor[A]]): ConfigDescriptor[List[A]] =
-    configList.foldLeft(Succeed(Nil): ConfigDescriptor[List[A]])(
+    configList.foldLeft(Empty[List[A]]().xmap(_.toList.flatten)(_.headOption.map(List(_))))(
```
afsalthaj: Is it same as 
“Empty[A]().xmap(_.toList)(_.headOption)?”

afsalthaj: May be I am wrong. But that should give a Config[List[A]] in return?

wi101: this will return `List[Option[Config]]` this is why I did `flatten`

wi101: I will use `fold` to make it clearer

PR #64, opened 2019-10-04
Add default

afsalthaj:


afsalthaj:
This PR becoming big.
It wil consist of solving `docs` to use a better tree type.

PR #65, opened 2019-10-05
Use System to access env and properties

Michaelt293:
In this pull request, `System` is used to access environment variables and system properties.

Note: this may be advantageous since `... even if the security manager does not permit the getProperties operation, it may choose to permit the getProperty(String) operation.` Also, this change should improve testability via environmental effects.

### DIFF
--- a/zio-config/src/main/scala/zio/config/Config.scala
+++ b/zio-config/src/main/scala/zio/config/Config.scala
@@ -3,7 +3,8 @@ package zio.config
 import java.net.URI
 
 import zio.config.actions.Read
-import zio.{ system, IO, Task, UIO, ZIO }
+import zio.system.System
+import zio.{ IO, RIO, UIO, ZIO }
 
 trait Config[A] {
   def config: Config.Service[A]
@@ -31,20 +32,20 @@ object Config {
     override def config: UIO[A] = ZIO.succeed(a)
   }
 
-  def fromEnv[A](configDescriptor: ConfigDescriptor[A]): ZIO[system.System, Throwable, Config[A]] =
+  def fromEnv[A](configDescriptor: ConfigDescriptor[A]): RIO[System, Config[A]] =
     for {
-      lineSep <- ZIO.accessM[zio.system.System](sys => sys.system.lineSeparator)
-      source  <- Task(sys.env).map(mapSource)
+      lineSep <- ZIO.accessM[System](_.system.lineSeparator)
+      source  <- envSource
       res     <- make(source, configDescriptor).mapError(t => new RuntimeException(t.mkString(lineSep)))
     } yield res
 
   def fromMap[A](map: Map[String, String], configDescriptor: ConfigDescriptor[A]): IO[ReadErrors, Config[A]] =
     make(mapSource(map), configDescriptor)
 
-  def fromPropertyFile[A](configDescriptor: ConfigDescriptor[A]): ZIO[system.System, Throwable, Config[A]] =
+  def fromPropertyFile[A](configDescriptor: ConfigDescriptor[A]): RIO[System, Config[A]] =
     for {
-      lineSep <- ZIO.accessM[zio.system.System](sys => sys.system.lineSeparator)
-      source  <- Task(sys.props.toMap).map(mapSource)
+      lineSep <- ZIO.accessM[System](_.system.lineSeparator)
+      source  <- propSource
       res     <- make(source, configDescriptor).mapError(t => new RuntimeException(t.mkString(lineSep)))
     } yield res
 
--- a/zio-config/src/main/scala/zio/config/Sources.scala
+++ b/zio-config/src/main/scala/zio/config/Sources.scala
@@ -1,39 +1,52 @@
 package zio.config
 
 import zio.system.System
-import zio.{ Task, ZIO }
+import zio.{ IO, ZIO }
 
 trait Sources {
-  def envSource: ZIO[System, Throwable, ConfigSource] =
+  val envSource: ZIO[System, Nothing, ConfigSource] =
     ZIO.access { env =>
       new ConfigSource {
         val configService: ConfigSource.Service =
           (path: String) =>
-            env.system.env(path).flatMap { (oStr: Option[String]) =>
-              oStr.fold[Task[String]](
-                ZIO.fail(new IllegalArgumentException(s"Not defined in system environment: $path"))
-              )(ZIO.succeed)
-            }
+            env.system
+              .env(path)
+              .mapError { err =>
+                ReadError.FatalError(path, err)
+              }
+              .flatMap { (oStr: Option[String]) =>
+                oStr.fold[IO[ReadError, String]](
+                  ZIO.fail(ReadError.MissingValue(path, "Not defined in system environment"))
+                )(ZIO.succeed)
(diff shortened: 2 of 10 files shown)


afsalthaj:
I will take a look at it. This is fantastic Michael

afsalthaj:
The idea of using zio.System seem
Native and natural as well. When I tried , it ended up in re-evaluation of config everytime we need to access it. Feel free to keep giving a hand on this.

zio-config/src/main/scala/zio/config/Config.scala
```diff
@@ -31,20 +32,20 @@ object Config {
     override def config: UIO[A] = ZIO.succeed(a)
   }
 
-  def fromEnv[A](configDescriptor: ConfigDescriptor[A]): ZIO[system.System, Throwable, Config[A]] =
+  def fromEnv[A](configDescriptor: ConfigDescriptor[A]): RIO[System, Config[A]] =
     for {
-      lineSep <- ZIO.accessM[zio.system.System](sys => sys.system.lineSeparator)
-      source  <- Task(sys.env).map(mapSource)
+      lineSep <- ZIO.accessM[System](_.system.lineSeparator)
+      source  <- envSource
```
afsalthaj: 👍

zio-config/src/main/scala/zio/config/Config.scala
```diff
@@ -31,20 +32,20 @@ object Config {
...
       res     <- make(source, configDescriptor).mapError(t => new RuntimeException(t.mkString(lineSep)))
     } yield res
 
   def fromMap[A](map: Map[String, String], configDescriptor: ConfigDescriptor[A]): IO[ReadErrors, Config[A]] =
     make(mapSource(map), configDescriptor)
 
-  def fromPropertyFile[A](configDescriptor: ConfigDescriptor[A]): ZIO[system.System, Throwable, Config[A]] =
+  def fromPropertyFile[A](configDescriptor: ConfigDescriptor[A]): RIO[System, Config[A]] =
```
afsalthaj: `RIO` 👍

zio-config/src/main/scala/zio/config/Config.scala
```diff
@@ -31,20 +32,20 @@ object Config {
...
     for {
-      lineSep <- ZIO.accessM[zio.system.System](sys => sys.system.lineSeparator)
-      source  <- Task(sys.props.toMap).map(mapSource)
+      lineSep <- ZIO.accessM[System](_.system.lineSeparator)
+      source  <- propSource
```
afsalthaj: Reuse of functionalities. And that's perfect

afsalthaj:
@Michaelt293 
I didn't know this point;
even if the security manager does not permit the getProperties operation, it may choose to permit the getProperty(String) operation.

Good one.

zio-config/src/main/scala/zio/config/Sources.scala
```diff
@@ -1,17 +1,39 @@
...
-  def propSource: Task[ConfigSource] =
-    Task(sys.props.toMap).map(mapSource)
+  def propSource: ZIO[System, Throwable, ConfigSource] =
+    ZIO.access { env =>
+      new ConfigSource {
+        val configService: ConfigSource.Service =
+          (path: String) =>
+            env.system.property(path).flatMap { (oStr: Option[String]) =>
+              oStr.fold[Task[String]](
+                ZIO.fail(new IllegalArgumentException(s"System property not found for key: $path"))
```
afsalthaj: @Michaelt293 

1. (s"System property not found for key: $path")). How is this different to MissingValue handling done in `Read.scala`. Could you take a look at it. You may have some inputs there.

```scala
            value <- config
                      .getConfigValue(path)
                      .mapError(_ => ReadErrors(MissingValue(path)))
```

2. This might be a silly question, but just to be on the same page. If we have `n` keys to read, then we call `sys.env(key)` for sure. And you are saying that's ok, because `sys.properties` may faill due to security, but `sys.property(key)` won't . Is that right?

afsalthaj: The only minor issue here, is 
```
           value <- config
                      .getConfigValue(path)
                      .mapError(_ => ReadErrors(MissingValue(path)))
```
We are mapping all errors `getConfigValue` which is now returning a `Task`, to be `MissingValue`.
Is there way you propagate this information. That is, ensuring that the failure in `Task` is actually due to a `MissingValue` and hence tagging it?
`.mapOnlyMissingValueError(error => MissingValue)`

afsalthaj: And try running an example giving `sys.env`. We don't have a test which actually test system environment.

afsalthaj: I think if this is in place, its good to merge 👍

afsalthaj: By doing thee above changes I mentioned, we get rid of illegal argument exceptions in the code

zio-config/src/main/scala/zio/config/package.scala
```diff
@@ -11,7 +11,8 @@ package object config extends Sources {
 
   def config[A]: ZIO[Config[A], Nothing, A] = ZIO.accessM(_.config.config)
 
-  def getConfigValue(path: String): ZIO[ConfigSource, Unit, String] = ZIO.accessM(_.configService.getConfigValue(path))
+  def getConfigValue(path: String): RIO[ConfigSource, String] =
+    ZIO.accessM(_.configService.getConfigValue(path))
```
afsalthaj: 👍

zio-config/src/main/scala/zio/config/Sources.scala
Michaelt293: Good points, I have used `ReadError` as the error type now. I have also added a `cause` field to `MissingValue` to provide more information on what went wrong.  I have also added a `FatalError` case to the `ReadError` type which ends up handling the `SecurityException`  that can occur when accessing environment and system properties.

 Regarding `sys.properties` vs `sys.property(key)`, `sys.property(key)` can still throw a `SecurityException`. My understanding is that things can be configured so that `sys.properties` could throw an exception and `sys.property(key)` returns a value. This is probably and edge case that you would typically would not encounter.

Michaelt293: Regarding testing, I have not manually tested these changes. Perhaps we should first try  this before merging this PR.

zio-config/src/main/scala/zio/config/Sources.scala
```diff
@@ -1,17 +1,52 @@
 package zio.config
 
-import zio.{ IO, Task }
+import zio.system.System
+import zio.{ IO, ZIO }
 
 trait Sources {
-  def envSource: Task[ConfigSource] =
-    Task(sys.env).map(mapSource)
+  val envSource: ZIO[System, Nothing, ConfigSource] =
```
Michaelt293: Note the use of `val` here

afsalthaj:
I think we can merge this. It's in that state.

afsalthaj:
LGTM.

We will be rethinking about ConfigSource having to know about ReadError. However, it's debatable. If at all we need to change, it's a very minor change.

PR #71, opened 2019-10-06
Rework ProgramExample for realistic bootstrapping of Spark etc

leigh-perry:


examples/src/main/scala/zio/config/examples/ProgramExample.scala
```diff
@@ -19,77 +20,108 @@ object ProgramExample extends App {
...
+  def sql(value: String): DataFrame =
+    new DataFrame {}
+
+  def version: String =
+    "someVersion"
+}
+
+trait DataFrame
+
+////
```
afsalthaj: Let's remove `///` as we are not using anywhere else in the code.

examples/src/main/scala/zio/config/examples/ProgramExample.scala
```diff
@@ -19,77 +20,108 @@ object ProgramExample extends App {
...
+              new Service {
+                override def sparkEnv: UIO[SparkSession] =
+                  ZIO.succeed(sparkSession)
+              }
+          }
+      )
+
+  def local(name: String): ZIO[Blocking, Throwable, SparkEnv] =
+    make {
+      // SparkSession.builder().appName(name).enableHiveSupport().getOrCreate()
```
afsalthaj: Add a comment,
An example , that it is an example of real world spark initiation.

examples/src/main/scala/zio/config/examples/ProgramExample.scala
```diff
@@ -19,77 +20,108 @@ object ProgramExample extends App {
...
+            println(
+              s"Executing something with programConfig's parameters ${r.inputPath} and ${r.outputPath} without the need of sparkSession"
             )
+          )
+    } yield ()
+
+  val runSparkJob: ZIO[SparkEnv with Blocking, Throwable, Unit] =
+    for {
+      session <- ZIO.accessM[SparkEnv](_.spark.sparkEnv)
+      result  <- zio.blocking.effectBlocking(session.sql("SELECT something"))
```
afsalthaj: session.sql returns a data frame, and it doesn't need to be blocking.
Have a write function and make it blocking

afsalthaj:
LGTM.
If you could remove effect blocking for returning data frame, and use it to write data instead, that will be good.

examples/src/main/scala/zio/config/examples/ProgramExample.scala
afsalthaj: Let's also remove `DataFrame`. It's a bit too much detail of spark in this example. Those who are unfamiliar with spark may find it a bit confusing.,

examples/src/main/scala/zio/config/examples/ProgramExample.scala
afsalthaj: Also `zio.blocking.effectBlocking(session.write("somedata"))`

examples/src/main/scala/zio/config/examples/ProgramExample.scala
```diff
@@ -19,77 +20,108 @@ object ProgramExample extends App {
...
+    }
   }
 }
 
-// spark service
+//// stubs for the real Spark
+
+trait SparkSession {
+  def sql(value: String): DataFrame =
+    new DataFrame {}
```
afsalthaj: `def write(value: String): Unit = ()`

examples/src/main/scala/zio/config/examples/ProgramExample.scala
```diff
@@ -19,77 +20,108 @@ object ProgramExample extends App {
...
+    for {
+      session <- ZIO.accessM[SparkEnv](_.spark.sparkEnv)
+      result  <- zio.blocking.effectBlocking(session.sql("SELECT something"))
+      _       <- ZIO.effect(println(s"Executing something with spark ${session.version}: ${result}"))
+    } yield ()
+
+  val processData: ZIO[SparkEnv with Config[ProgramConfig], Throwable, Unit] =
+    for {
+      conf  <- config[ProgramConfig]
+      spark <- ZIO.access[SparkEnv](_.spark)
```
afsalthaj: Why do u need to get spark explicitly?

afsalthaj: `ZIO.accessM[SparkEnv](env => effectBlocking(env.write("message"))` is enough

examples/src/main/scala/zio/config/examples/ProgramExample.scala
```diff
@@ -19,77 +20,108 @@ object ProgramExample extends App {
...
     } yield ()
 
-  val execute: ZIO[SparkEnv with Config[ProgramConfig], Throwable, Unit] =
+  val execute: ZIO[SparkEnv with Config[ProgramConfig] with Blocking, Throwable, Unit] =
     for {
-      _ <- logSparkSession
       _ <- logProgramConfig
-      _ <- processData
+        _ <- runSparkJob
+        _ <- processData
```
afsalthaj: I think this for-comprehension isn't formatted.

examples/src/main/scala/zio/config/examples/ProgramExample.scala
leigh-perry: `ZIO.access[SparkEnv](_.spark)` was a bug, hidden because the example only used its `toString`. Should have been `ZIO.accessM[SparkEnv](_.spark)`

PR #72, opened 2019-10-06
Multiple improvements

afsalthaj:
1. Remove more than 2 aliases for functions if there is any
2. Change from `~` to `?` for docs.
3. `read` will return only `A` now, and not `(ConfigReport, A)`. We separated out this concern to `docs`. `ConfigReport` is deleted as such.
5. Make the function names standard one based on the core zio library.
6. Implement `Config.Default`. Now you can do `string("URL").default("localhost")`
7. Remove `OnError`. 
8. Come up with a better structure (a tree now, that is recursive) to express the config docs. This is much better expressive than before, and can also accept `docs(desc, Some(a))` and docs(desc , None), while the former will emit the values held by each path. Have a look at the example usage below.

```scala
// Example
  final case class PgmConfig(a: String, b: Int)

  val pgmConf: ConfigDescriptor[PgmConfig] =
    (string("HELLO").default("xyz") |@|
      string("SOMETHING").orElseEither(int("PORT").default(1))
      )(PgmConfig.apply, PgmConfig.unapply)

  assert(
    docs(pgmConf, Some(result)) ==
     And(
        Leaf(KeyDescription("HELLO", Some("xyz"), List("value of type string", "default value: xyz"))),
        Or(
          Leaf(KeyDescription("SOMETHING", None, List("value of type string"))),
          Leaf(KeyDescription("PORT", Some("1"), List("value of type int", "default value: 1")))
        )
      )
  )
```
9. A better package object, however, this is still in question how much more code has to be removed. Still need a discussion on this.
10. ProductBuilder.scala code generation  is made part of sbt task. This will not run everytime. User can run `geenrateProductBuilder` if making any change in local. 
It overrides the scalafmt as well to look better. it supports 22 now.
11. Removed an unnecessary `asInstanceOf`
12. Removed all implicits


zio-config/src/main/scala/zio/config/actions/Read.scala
```diff
@@ -1,94 +1,86 @@
...
-            case (r, src) => ZIO.fromEither(f(src)).bimap(err => ReadErrors(err), res => (r, res))
+        case ConfigDescriptor.XmapEither(c, f, _) =>
+          loop(c, previousDescription).flatMap { a =>
+            ZIO.fromEither(f(a)).bimap(err => ReadErrors(err), res => res)
           }
 
+        // No need to add report on the default value.
+        case ConfigDescriptor.Default(c, value) =>
+          loop(c, previousDescription).fold(
+            _ => value,
```
Michaelt293: I think it would be better to inspect the error here. If the value is missing, the default value should be supplied. If there is a parse error or some fatal error, we should report the error (i.e., not supply a default value).

afsalthaj: Yes. I think it makes sense

README.md
```diff
@@ -32,16 +32,17 @@ case class Prod(ldap: String, port: Int, dburl: Option[String])
...
     for {
-      prod    <- config[Prod]
-      written <- write(Prod.prodConfig).run.provide(prod)
-      _       <- ZIO.effect(println(written))
-      _       <- ZIO.effect(println(userManual(Prod.prodConfig)))
+      prodConf <- config[Prod]
+      written  <- write(Prod.prodConfig).run.provide(prodConf)
+      report    = docs(Prod.prodConfig, Some(prod))
+      _        <- ZIO.effect(println(written))
+      _        <- ZIO.effect(println(report))
```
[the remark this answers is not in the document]
afsalthaj: Done!

zio-config/src/main/scala/zio/config/actions/Read.scala
afsalthaj: I will take this in follow up PR.

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
```diff
@@ -1,49 +1,100 @@
...
 
-final case class ConfigDocs(configKeysAndDescription: List[KeyDescription], or: Option[ConfigDocs])
+sealed trait ConfigDocs
 
 // Man page
 object ConfigDocs {
-  final case class KeyDescription(path: String, list: List[String])
 
-  final def docs[A](config: ConfigDescriptor[A]): ConfigDocs = {
+  final case class Empty()                                  extends ConfigDocs
```
wi101: You can make `Empty` an `object`

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
```diff
@@ -1,49 +1,100 @@
...
+  final case class Leaf(value: KeyDescription)              extends ConfigDocs
+  final case class And(left: ConfigDocs, right: ConfigDocs) extends ConfigDocs
+  final case class Or(left: ConfigDocs, right: ConfigDocs)  extends ConfigDocs
+
+  final case class KeyDescription(path: String, value: Option[String], docs: List[String])
```
wi101: the `value` is the description?

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
```diff
@@ -1,49 +1,100 @@
...
+
+  final def createDoc[A](config: ConfigDescriptor[A], value: Option[A]): ConfigDocs = {
     def loop[B](
-      acc: List[String],
+      accumulatedDoc: List[String],
       previousDescription: String,
       config: ConfigDescriptor[B],
-      desc: ConfigDocs
+      desc: ConfigDocs,
+      value: Option[B]
```
wi101: is this a trick to avoid `asInstanceOf` in case of `Optional`?

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
afsalthaj: The value is the actual value of the path. User may or may not have given a value of the path. In that case we say the value is None and still provide a description

afsalthaj: If you could take a look at the example, it will be much clear

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
afsalthaj: No. The asInstanceOf was existing for no reason. The option is because for a given key user may or may not have passed a value.

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
afsalthaj: Sure

zio-config/src/main/scala/zio/config/actions/Read.scala
```diff
@@ -1,94 +1,86 @@
...
-
-        case ConfigDescriptor.OnError(c, f) =>
-          ZIO.accessM[ConfigSource](
-            _ =>
-              loop(c, report, previousDescription)
-                .foldM(
-                  errors => report.get.map(r => (r, f(errors))),
-                  UIO(_)
-                )
+          loop(c, previousDescription).fold(
```
wi101: consider to use `.option`

afsalthaj: Sure.

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
wi101: I see, Thanks :)

afsalthaj:
Thanks Wiem for the review

examples/src/main/scala/zio/config/examples/ReadWriteReport.scala
```diff
@@ -2,91 +2,89 @@ package zio.config.examples
...
-      (string("xyz") |@| int("abc"))(XYZ.apply, XYZ.unapply).optional ~ "Example: xyz")(
+    ((string("usr") ? "Example: some-user" |@|
+      string("pwd").xmap(Password)(_.value).optional ? "sec" |@|
+      string("jhi").optional ? "Ex: ghi" |@|
+      (string("xyz") |@| int("abc").orElseEither(string("def")))(XYZ.apply, XYZ.unapply).optional ? "Ex: ha")(
       UserPwd.apply,
       UserPwd.unapply
-    ) or
-      (string("auth_token") |@| string("clientid"))(Token.apply, Token.unapply)) ~ "Prod Config"
+    ) orElseEither
```
jdegoes: I think we should introduce `<+>` as the alias for `orElseEither` (it being the dual of `<*>` / `zip`).

zio-config/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -1,40 +1,55 @@
...
     self.xmapEither(a => Right(to(a)))(b => Right(from(b)))
 
-  def xmapEither2[B, C](
-    that: ConfigDescriptor[B]
-  )(f: (A, B) => Either[ReadError, C])(g: C => Either[String, (A, B)]): ConfigDescriptor[C] =
-    (self |@| that).apply[(A, B)]((a, b) => (a, b), t => Some((t._1, t._2))).xmapEither(b => f(b._1, b._2))(g)
+  final def orElseEither[B](that: => ConfigDescriptor[B]): ConfigDescriptor[Either[A, B]] =
+    ConfigDescriptor.OrElseEither(self, that)
+
+  final def <+>[B](that: => ConfigDescriptor[B]): ConfigDescriptor[Either[A, B]] =
```
jdegoes: Ah, I see we already have it! 😆

afsalthaj: Yea :) 👍

PR #76, opened 2019-10-08
Simplify write

afsalthaj:
Addressing previous review comments.
Simplify write to use raw ZIO.


zio-config/src/main/scala/zio/config/actions/Read.scala
```diff
@@ -45,10 +45,7 @@ object Read {
           loop(c, message)
 
         case ConfigDescriptor.Optional(c) =>
-          loop(c, previousDescription).fold(
-            _ => None,
-            success => Some(success)
-          )
+          loop(c, previousDescription).option
```
leigh-perry: Nice :-)

PR #78, opened 2019-10-09
Nested Read, Write and Documentation.

afsalthaj:


zio-config/src/main/scala/zio/config/actions/Read.scala
```diff
@@ -1,25 +1,30 @@
...
 
         case ConfigDescriptor.Source(path, propertyType) =>
           for {
             value <- config
                       .getConfigValue(path)
-                      .mapError(err => ReadErrors(err))
+                      .flatMap({
+                        case PropertyTree.Leaf(value) => ZIO.succeed(value)
+                        case PropertyTree.Record(_) =>
+                          ZIO.fail(ReadError.FoundNestedObject(path): ReadError[String, String])
```
afsalthaj: Once we have a new node, `Nested`, the `path` will be `List[Path]` or something that goes from `List[Path] => String` that can fetch the value deep in the tree.

`.get("database.connection.url")`

zio-config/src/main/scala/zio/config/package.scala
```diff
@@ -16,16 +16,21 @@ package object config extends Sources {
...
-    def apply(a: ReadError, as: ReadError*): ReadErrors =
+    def apply[K, V](a: ReadError[K, V], as: ReadError[K, V]*): ReadErrors[K, V] =
       ::(a, as.toList)
 
-    def concat(l: ReadErrors, r: ReadErrors): ReadErrors =
+    def concat[K, V](l: ReadErrors[K, V], r: ReadErrors[K, V]): ReadErrors[K, V] =
       ::(l.head, l.tail ++ r)
+
+    def asThrowable[K, V, A](readErrors: ReadErrors[K, V]): RIO[system.System, A] =
+      zio.system.lineSeparator.flatMap(t => ZIO.fail(new RuntimeException(readErrors.mkString(t))))
```
afsalthaj: A helper for the user in case they want to convert to throwable

Michaelt293: How do you feel about extending the error types by `Exception`?

```
sealed trait ReadError extends Exception {
  val key: String
}

  final case class ReadErrors(errors: ::[ReadError]) extends Exception {
    override def getMessage(): String = ???
  }

  object ReadErrors {
    def apply(a: ReadError, as: ReadError*): ReadErrors =
      ReadErrors(::(a, as.toList))

    def concat(l: ReadErrors, r: ReadErrors): ReadErrors =
      ReadErrors(::(l.errors.head, l.errors.tail ++ r.errors))
  }
```

afsalthaj:
Well. I guess it doesn’t need to be of the type exception. We need to give
user that flexibility. Staying away from exception sounding better to me

zio-config/src/main/scala/zio/config/package.scala
afsalthaj: I prefer errors not extending exception unless it really benefits us. Especially in this case, bootstrapping a config happens at the main resulting in some read errors. And we usually resolve it to folding over the result and logging the error  and zio.succeed(1)

zio-config/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -1,13 +1,52 @@
...
-  val configService: ConfigSource.Service
+trait ConfigSource[K, V] {
+  val configSourceService: ConfigSource.Service[K, V]
 }
 
 object ConfigSource {
-  trait Service {
-    def getConfigValue(path: String): IO[ReadError, String]
+  trait Service[K, V] {
+    def getConfigValue(path: K): IO[ReadError[K, V], PropertyTree[K, V]]
```
jdegoes: This is the "non-flattened" representation which may be more difficult to support for, e.g. system properties, environment variables, etc. 

The "flattened" representation is:

```scala
def getConfigValue(path: List[K]): IO[ReadError[K, V], V]
```

This representation permits the possibility that intermediate nodes cannot be read; only terminal nodes can be read. Which is true for completely flat sources with emulated nesting (e.g. `foo.bar.baz = 123`).

zio-config/src/test/scala/zio/config/ReadWriteRoundtripTest.scala
```diff
@@ -52,7 +52,7 @@ object ReadWriteRoundtripTest extends Properties("Coproduct support") with TestS
     val p2 =
       for {
         written <- write(cId).provide(p)
-        reread  <- read(cId).provide(mapSource(written))
+        reread  <- read(cId).provide(ConfigSource.mapSource(written))
```
jdegoes: I might rename `mapSource` to `fromMap`, etc., following the convention of CD.

zio-config/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -0,0 +1,11 @@
+package zio.config
+
+sealed trait PropertyTree[+K, +V]
```
jdegoes: I'd also add `flatten`, we can do lots others too, `mapKey`, `map` / `mapValue`, `fold`, etc.

afsalthaj:
Sure. Yes. I will be adding them.

afsalthaj:
More changes coming here. Will reopen this in a day

afsalthaj:
@jdegoes Could you re-review?
The main result is:
https://github.com/zio/zio-config/blob/c52c6db24ecef60169712f8be48c2f4b22d995f2/examples/src/main/scala/zio/config/examples/NestedConfigExample.scala

afsalthaj:
@Michaelt293 We can take whether to extend Excption or not in a follow up PR?
This PR is already a big one

zio-config/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -58,15 +60,17 @@ object ConfigDescriptor {
...
 
-  final case class Default[A](configDescriptor: ConfigDescriptor[A], value: A) extends ConfigDescriptor[A]
+  final case class Default[A](config: ConfigDescriptor[A], value: A) extends ConfigDescriptor[A]
 
   final case class Optional[A](config: ConfigDescriptor[A]) extends ConfigDescriptor[Option[A]]
 
   final case class XmapEither[A, B](
     config: ConfigDescriptor[A],
-    f: A => Either[ReadError, B],
+    f: A => Either[ReadError[String, String], B],
```
jdegoes: You could potentially push `K` and `V` into `ConfigDescriptor`, and make them covariant, but perhaps that's something we should look at later.

zio-config/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -0,0 +1,33 @@
...
+  final case object Empty extends PropertyTree[Nothing, Nothing]
+
+  final case class Leaf[V](value: V) extends PropertyTree[Nothing, V]
+
+  final case class Record[K, V](value: Map[K, PropertyTree[K, V]]) extends PropertyTree[K, V]
+
+  def fromMap[K, V](map: Map[K, V]): PropertyTree[K, V] =
+    Record(map.map(t => t._1 -> Leaf(t._2)))
+
+  def flattenKV[K, V](propertyTree: PropertyTree[K, V], zero: K, append: (K, K) => K): Map[K, V] = {
```
jdegoes: I'd put this as a method on `PropertyTree` so you can call it using OO syntax.

zio-config/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -0,0 +1,33 @@
...
+      propertyTree match {
+        case Empty         => acc
+        case Leaf(v)       => acc ++ Map(key -> v)
+        case Record(value) => value.flatMap(t => go(append(key, t._1), t._2, acc))
+      }
+
+    go(zero, propertyTree, Map.empty)
+  }
+
+  def flatten(propertyTree: PropertyTree[String, String], appendString: String = "."): Map[String, String] =
```
jdegoes: Ditto, method on `PropertyTree`, with implicit evidence that `K =:= String`, `V =:= String`.

zio-config/src/main/scala/zio/config/PropertyType.scala
```diff
@@ -7,73 +7,73 @@ import zio.config.ReadError.ParseError
 import scala.util.{ Failure, Success, Try }
 
 trait PropertyType[A] {
-  def read(path: String, propertyValue: String): Either[ReadError, A]
+  def read(path: String, propertyValue: String): Either[ReadError[String, String], A]
```
jdegoes: Yes, these must all be specialized to string here.

Of course, we can push `K` and `V` into `PropertyType`, and we probably should...

zio-config/src/main/scala/zio/config/ReadError.scala
```diff
@@ -1,11 +1,11 @@
...
+  val key: K
 }
 
 object ReadError {
-  final case class MissingValue(key: String)                                 extends ReadError
-  final case class ParseError(key: String, provided: String, `type`: String) extends ReadError
-  final case class FatalError(key: String, cause: Throwable)                 extends ReadError
+  final case class MissingValue[K](key: K)                               extends ReadError[K, Nothing]
+  final case class ParseError[K, V](key: K, provided: V, `type`: String) extends ReadError[K, V]
+  final case class FatalError[K](key: K, cause: Throwable)               extends ReadError[K, Nothing]
```
jdegoes: Did we end up needing this one?

zio-config/src/main/scala/zio/config/ReadError.scala
```diff
@@ -1,11 +1,11 @@
...
+sealed trait ReadError[+K, +V] {
+  val key: K
 }
 
 object ReadError {
-  final case class MissingValue(key: String)                                 extends ReadError
-  final case class ParseError(key: String, provided: String, `type`: String) extends ReadError
-  final case class FatalError(key: String, cause: Throwable)                 extends ReadError
+  final case class MissingValue[K](key: K)                               extends ReadError[K, Nothing]
+  final case class ParseError[K, V](key: K, provided: V, `type`: String) extends ReadError[K, V]
```
jdegoes: Maybe rename `type` => `message`?

zio-config/src/main/scala/zio/config/actions/Write.scala
```diff
@@ -1,52 +1,77 @@
...
-            ZIO.succeed(Map.empty[String, String])
-          )(aa => write(c).provide(aa))
-        )
-
-      case ConfigDescriptor.Default(c, _) =>
-        write(c)
-
-      case ConfigDescriptor.XmapEither(c, _, to) =>
-        ZIO.accessM { b =>
+  final def write[A](config: ConfigDescriptor[A], a: A): Either[String, PropertyTree[String, String]] = {
```
jdegoes: Ooh this looks so much better! ❤️

afsalthaj:
@jdegoes Thanks for feedback :)
Il probably come up with a follow up addressing a couple of more review comments. Good to get this merged now.

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
```diff
@@ -6,79 +6,82 @@ sealed trait ConfigDocs
...
 object ConfigDocs {
 
-  final case class Empty()                                  extends ConfigDocs
-  final case class Leaf(value: KeyDescription)              extends ConfigDocs
-  final case class And(left: ConfigDocs, right: ConfigDocs) extends ConfigDocs
-  final case class Or(left: ConfigDocs, right: ConfigDocs)  extends ConfigDocs
-
-  final case class KeyDescription(path: String, value: Option[String], docs: List[String])
+  final case class Empty()                                                                    extends ConfigDocs
+  final case class PathDetails(path: List[String], value: Option[String], docs: List[String]) extends ConfigDocs
```
wi101: Could you use `Vector` instead of `List` to take advantage of the optimization of `++` in `Vector` (in case you're using it)

afsalthaj:
@jdegoes @wi101 Thanks for the review. Made the changes

zio-config/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -1,13 +1,66 @@
...
+      val configSourceService: ConfigSource.Service[String, String] =
+        (path: List[String]) => {
+          val key = path.mkString(delimiter)
+          ZIO.fromOption(map.get(key)).mapError { _ =>
+            ReadError.MissingValue(key): ReadError[String, String]
+          }
+        }
+    }
+
+  def fromPropertyTree(
```
jdegoes: Maybe: `fromPropertyTreeString`? At the risk of making method names too long. 😆 

In theory, we could have `fromPropertyTree[K, V]` at some point.

zio-config/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -0,0 +1,36 @@
...
+      propertyTree match {
+        case Empty         => acc
+        case Leaf(v)       => acc ++ Map(key -> v)
+        case Record(value) => value.flatMap(t => go(append(key, t._1), t._2, acc))
+      }
+
+    go(zero, self, Map.empty[K1, V1])
+  }
+
+  def flatten[K1 >: K, V1 >: V](
```
jdegoes: I'd call this one `flattenString`

zio-config/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -0,0 +1,36 @@
+package zio.config
+
+import zio.config.PropertyTree.{ Empty, Leaf, Record }
+
+sealed trait PropertyTree[+K, +V] { self =>
+  def flattenKV[K1 >: K, V1 >: V](zero: K1, append: (K1, K1) => K1): Map[K1, V1] = {
```
jdegoes: I'd call this one `flatten`

PR #79, opened 2019-10-10
Use lineSeparator defined in package object system

Michaelt293:
Pretty minor change, we can use `lineSeparator` defined in the `system` package object -
https://github.com/zio/zio/blob/06d3257a2a58aed7721c0de2b0422f27f07fd491/core/shared/src/main/scala-2.12%2B/zio/system.scala


afsalthaj:
@Michaelt293 I have taken off using lineSeparator and RuntimeException usage in latest PR #78 Would you mind closing this PR?

PR #80, opened 2019-10-11
WIP: Combine sources

Michaelt293:
This pull request adds support for combining config sources. A `sourceDescription` has been added to `ConfigSource` service. This should allow more information to be passed to the `ConfigDocs` with some re-factoring  of the current code base.


Michaelt293:
@afsalthaj  I have added the methods directly to the `ConfigSource` trait and removed the use of an implicit class. In addition, I have modified the `getConfigValue` method to have the return type `IO[ReadError, Value]` where `Value` is a case class containing both the value and source description. The changes should allow us to combine multiple sources and report from which source the value was obtained from (as well as give a full description of the combine config source). To do this, we will need to use the `ConfigSource`  when creating the `ConfigDocs`. Also, I believe we should create the `ConfigDocs`  when we instantiate our config type (`A` in `read`) since reading values from an external source is inherently stateful. Therefore, I believe the type of `read` should be -

`def read[A](configuration: ConfigDescriptor[A]): ZIO[ConfigSource, ReadErrors, (ConfigDocs, A)]`

I know that the return type of `read` was recently changed and agree that separating the generation of config docs and instantiating the config type is certainly cleaner. Due to the reasons above, I can't see a better way to make this work. I am certainly interested in your ideas :)

zio-config/src/main/scala/zio/config/Sources.scala
```diff
@@ -8,37 +8,54 @@ trait Sources {
...
+          new ConfigSource.Service {
+            val sourceDescription: String = "Environment"
+
+            def getConfigValue(path: String): IO[ReadError, ConfigSource.Value] =
+              env.system
+                .env(path)
+                .mapError { err =>
+                  ReadError.FatalError(path, err)
+                }
+                .flatMap(IO.fromOption(_).mapError(_ => ReadError.MissingValue(path)))
```
wi101: you can use foldM instead
```scala
IO.fromOption(_).foldM(IO.fail(ReadError.MissingValue(path)), UIO(ConfigSource.Value(value, sourceDescription)))
```

zio-config/src/main/scala/zio/config/Sources.scala
```diff
@@ -8,37 +8,54 @@ trait Sources {
...
+            val sourceDescription: String = "System properties"
+
+            def getConfigValue(path: String): IO[ReadError, ConfigSource.Value] =
+              env.system
+                .property(path)
+                .mapError { err =>
+                  ReadError.FatalError(path, err)
+                }
+                .flatMap(IO.fromOption(_).mapError(_ => ReadError.MissingValue(path)))
+                .map(value => ConfigSource.Value(value, sourceDescription))
```
wi101: the same here

zio-config/src/main/scala/zio/config/actions/Read.scala
```diff
@@ -23,7 +23,7 @@ object Read {
 
             result <- ZIO.fromEither(
                        propertyType
-                         .read(path, value)
+                         .read(path, value.value)
```
wi101: `value.value` sounds quite confused, maybe we can rename the first `value` to something like `configV`or `configValue`

afsalthaj:
@Michaelt293 i understand why we ended up having a tuple back during the read . However it is a bit unfortunate we will have to bring them
back as a tuple. It was a design decision to not have it. There is a bit of abstraction leakage somewhere that we are having that coupling. 
@jdegoes any inputs on this? Will need your help on this. .

zio-config/src/main/scala/zio/config/actions/Read.scala
Michaelt293: Agree, replaced with `configValue`.

zio-config/src/main/scala/zio/config/Sources.scala
Michaelt293: Thanks, done :)

Michaelt293:
I understand, what about having to methods? One called `read` which returns the `A` and another call `readWithConfigDocs` which returns the tuple. Or perhaps using a similar idea elsewhere.

zio-config/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -2,12 +2,76 @@ package zio.config
...
+              .catchSome {
+                pf.andThen(
+                  _.configService
+                    .getConfigValue(path)
+                )
+              }
+        }
+    }
+
+  final def ifMissingValue(that: => ConfigSource): ConfigSource =
```
jdegoes: How does this differ from `orElse`?

Also it should probably be named `orElseXYZ`

zio-config/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -2,12 +2,76 @@ package zio.config
...
+                that.configService
+                  .getConfigValue(path)
+              )
+        }
+    }
+
+  final def <>(that: ConfigSource): ConfigSource =
+    orElse(that)
+
+  final def catchSome(
```
jdegoes: This doesn't have the idiomatic signature of `catchSome`, so I'd recommend a new name. Also does it really make sense to have a single fallback? I can create a generic signature here if you can explain to me the purpose of the method better.

zio-config/src/main/scala/zio/config/ConfigSource.scala
wi101: with the current implementation, this method will take the config `that` only for a specific Error (MissingValue)
would it make sense if we choose `that` when the current config fails and we could rename it `orElse`
@Michaelt293 wdyt?

wi101:
@Michaelt293 please consider to adjust the code and update the branch
Thanks!

zio-config/src/main/scala/zio/config/ConfigSource.scala
wi101: Ah well we arlready have `orElse` ...

Michaelt293: Yes, that's the reason. How about the name `orElseMissing`? Not really sure if that is clear enough. More informative names could get too long, e.g.,  `orElseIfMissingValue`.

zio-config/src/main/scala/zio/config/ConfigSource.scala
Michaelt293: That makes sense, I can change the name of the method. The purpose of this method is to allow users to build their own "fallback" methods if required. For example, a user may define a new config source and may want finer control when falling back on another config source. This method is probably not required so for simplicity, maybe it is best to remove this method.

zio-config/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -8,6 +8,64 @@ trait ConfigSource {
 
 object ConfigSource {
   trait Service {
+    val sourceDescription: String
+
     def getConfigValue(path: String): IO[ReadError, String]
   }
+
+  implicit class ConfigSourceOps(private val source: ConfigSource) extends AnyVal {
+    def orElse(that: ConfigSource): ConfigSource =
```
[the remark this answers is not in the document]
Michaelt293: Sure, this use of implicits has been removed.

afsalthaj:
I remember getting excited with sourceDescription. However if it isn’t used
then may be remove?

zio-config/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -8,6 +8,64 @@ trait ConfigSource {
 
 object ConfigSource {
   trait Service {
+    val sourceDescription: String
```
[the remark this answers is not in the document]
Michaelt293: Sure, if delaying merging this to master is better, go for it.

Michaelt293:
@wi101 @jdegoes Thanks for your review, will update this PR.

Michaelt293:
> I remember getting excited with sourceDescription. However if it isn’t used then may be remove?

See my previous discussion (in short I want to use it, still WIP :) ) -

> I have added the methods directly to the ConfigSource trait and removed the use of an implicit class. In addition, I have modified the getConfigValue method to have the return type IO[ReadError, Value] where Value is a case class containing both the value and source description. The changes should allow us to combine multiple sources and report from which source the value was obtained (as well as give a full description of the combined config source). To do this, we will need to use the ConfigSource when creating the ConfigDocs. Also, I believe we should create the ConfigDocs when we instantiate our config type (A in read) since reading values from an external source is inherently stateful. Therefore, I believe the type of read should be -
> 
> def read[A](configuration: ConfigDescriptor[A]): ZIO[ConfigSource, ReadErrors, (ConfigDocs, A)]
> 
> I know that the return type of read was recently changed and agree that separating the generation of config docs and instantiating the config type is certainly cleaner. Due to the reasons above, I can't see a better way to make this work. I am certainly interested in your ideas :)
>

afsalthaj:
@Michaelt293 Sure. 
> I know that the return type of read was recently changed and agree that separating the generation of config docs and instantiating the config type is certainly cleaner. Due to the reasons above, I can't see a better way to make this work. I am certainly interested in your ideas :)

I clearly see the point. I think its a trade off, of how much cleaner vs adding-more-functionality.

Another alternative, which I think you mentioned is having`read`, and `readWithConfigDocs`.  After thinking about it for a while, I think, it ends up in too much of things for the user - for a configuration library. Furthermore, another `docs` function may make it much more confusing, however, removing `docs` isn't a good idea at this stage - it's standalone and we find it really useful -it act as a man page (not just reporting ) that doesn't need a source !

Hmm, another alternative is, let the user do it, with the ZIO effect system , and they could use `orElse` or `orElseEither` directly.  The latter gives the user which source as well.

```
 read(desc).provide(source1).orElseEither(read(config).provide(source2))

```
If the source has description, then the user can use it to log the fact that X comes from source A.

Michaelt293:
Thanks for your comments. I understand, we have to keep the right balance and not end up with too much complexity. I can continue to work on this and we can see whether it's a worthwhile change. If it isn't, no problems.

afsalthaj:
Also, please be sure to be in sync with master, so that you don't have hard time resolving conflicts. Just that there are always significant code movements / renamings etc

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
```diff
@@ -6,10 +6,16 @@ sealed trait ConfigDocs
...
-  final case class And(left: ConfigDocs, right: ConfigDocs)                                     extends ConfigDocs
-  final case class Or(left: ConfigDocs, right: ConfigDocs)                                      extends ConfigDocs
+  final case class Empty() extends ConfigDocs
+  final case class PathDetails(
+    path: Vector[String],
+    value: Option[String],
+    source: Option[String],
+    docs: List[String]
+  ) extends ConfigDocs
+  final case class SourceDescription(sourceDescription: String, configDocs: ConfigDocs) extends ConfigDocs
```
afsalthaj: A bit confused here. How is this node conceptually different from the source parameter in PathDetails?

afsalthaj:
Let’s take this PR to the design discussions prior to the release. It’s worth discussing. I Reckon an application Not being certain about the source of a particular  config value  is a complicated situation. I would personally avoid that situation. 

If user has different configurations to be fetched from different sources they could explicitly get it from different sources by providing the source one by one. This makes it easier for user to reason about. Example their zio environment will then become 

ConfigEnv[MyConigFromMap] with ConfigEnv[MyConfigFromDatabase] with ConfigEnv[MyConfigFromPropertyFile]

I understand the importance of fallback. However having fallbacks could be a bit of trouble and if it all it’s a legit situation, isn’t it still a corner case, which can handle anyways using orElse of zio directly. 

However, user could choose to have multiple sources to form the configs in different parts of the application, and that can be handled with what I mentioned above. There is a subtle difference between not being certain about source/ fallback and handling multiple sources that we fully know.

zio-config/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -2,30 +2,115 @@ package zio.config
...
 import zio.config.ReadErrors.ReadError
 import zio.system.System
-import zio.{ IO, ZIO }
+import zio.{ IO, UIO, ZIO }
 
-trait ConfigSource[K, V] {
+trait ConfigSource[K, V] { self =>
   val configSourceService: ConfigSource.Service[K, V]
+
+  final def orElse(that: ConfigSource[K, V]): ConfigSource[K, V] =
```
jdegoes: I haven't looked at the implementation, but the operation makes perfect sense and is well-defined.

In addition, you could consider `<+>` / `orElseEither`, that allows the value types to vary (`V`, `V1`), returning `ConfigSource[K, Either[V, V1]]`.

zio-config/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -2,30 +2,115 @@ package zio.config
...
+                that.configSourceService
+                  .getConfigValue(path)
+              )
+        }
+    }
+
+  final def <>(that: ConfigSource[K, V]): ConfigSource[K, V] =
+    orElse(that)
+
+  final def catchMissingValue(that: => ConfigSource[K, V]): ConfigSource[K, V] =
```
jdegoes: This is the same as `orElse`, except it only looks for `MissingValue`, not all `ReadError`. This may be a fringe use case and I think there is a more orthogonal way to handle this, anyway... may be worth dropping for now.

zio-config/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -2,30 +2,115 @@ package zio.config
...
+                  that.configSourceService
+                    .getConfigValue(path)
+              }
+        }
+    }
+
+  final def <>?(that: ConfigSource[K, V]): ConfigSource[K, V] =
+    catchMissingValue(that)
+
+  final def map[V2](f: V => V2): ConfigSource[K, V2] =
```
jdegoes: Well defined, as is `mapKeys`.

zio-config/src/main/scala/zio/config/ConfigSource.scala
afsalthaj: When we say orElse, are we considering the error that happens  in forming the configsource or when fetching the value for a vector of Paths?

afsalthaj:
If you want to reconsider readWithConfigDocs, otherwise it’s all excellent to merge. (Also after fixing johns comments)

zio-config/src/main/scala/zio/config/ConfigSource.scala
jdegoes: The latter, I think, since if you have `UIO[ConfigSource[K, V]]` (a "builder" for `ConfigSource`) then you can already `orElse` that together using ZIO `orElse`, with another "builder".

afsalthaj: Makes sense :)

zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
Michaelt293: We can have primary sources, like a scala map or system properties, or a source composed of multiple source (such as system properties with a scala map as a fallback). `SourceDescription` describes the source as a whole, for example, `"Scala Map - south details: with fallback source: Scala Map - east details (on all errors)"`. The source parameter in `PathDetails` tells you precisely where the value was obtained (ie, system properties, scala map or config descriptor).

zio-config/src/main/scala/zio/config/ConfigSource.scala
Michaelt293: Sure, I can drop this.

zio-config/src/main/scala/zio/config/actions/Read.scala
```diff
@@ -83,4 +83,104 @@ object Read {
...
+
+        // No need to add report on the default value.
+        case ConfigDescriptor.Default(c, value) =>
+          loop(c, paths, acc, docs).orElse(
+            ZIO.succeed(
+              (
+                value,
+                ConfigDocs.PathDetails(
+                  paths, // empty vector?
+                  None,  // change to Some(value.toString)?
```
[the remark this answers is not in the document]
Michaelt293: Yes, the handling of default values was incorrect before. I have fixed this issue now.

zio-config/src/main/scala/zio/config/ConfigSource.scala
Michaelt293: I can add `mapKey`.

zio-config/src/main/scala/zio/config/actions/Read.scala
```diff
@@ -83,4 +83,104 @@ object Read {
...
+          }
+
+        // No need to add report on the default value.
+        case ConfigDescriptor.Default(c, value) =>
+          loop(c, paths, acc, docs).orElse(
+            ZIO.succeed(
+              (
+                value,
+                ConfigDocs.PathDetails(
+                  paths, // empty vector?
```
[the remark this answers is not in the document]
Michaelt293: Absolutely, this has now been resolved.

Michaelt293:
I have some ideas on `read` and `createDoc`. I have done some refactoring but I think there could still be some bugs in the recursive function I have worked on. Once I have got this to a state I am happy with, I'll remove the WIP.

afsalthaj:
Closing this for now. We will reopen once we start working on it again.

PR #106, opened 2019-11-03
Feature/source compositions

afsalthaj:


zio-config/src/main/scala/zio/config/actions/ConfigDocs.scala
```diff
@@ -1,85 +1,93 @@
...
+            loop(descAcc, left, docs),
+            loop(descAcc, right, docs)
+          )
+      }
+
+    loop(Descriptions(Nil), config, Empty())
+  }
+
+  def createDocWithValues[K, V, A](config: ConfigDescriptor[K, V, A], value: A): Either[String, ConfigDocs[K, V]] = {
+    val propertyTree = Write.write(config, value)
```
afsalthaj: I think Either makes makes sense. Given some value A, it doesn't necessarily mean, it can purely covert those values to its representation in a sys env / properties.

PR #126, opened 2019-12-01
Magnolia and micro site improvement

afsalthaj:
An initial version that works with Magnolia - automatic config description


afsalthaj:
@jdegoes  It doesn't work for scala-2.11, it seems.
However, could you review this initial PR on magnolia?

examples/src/main/scala/zio/config/examples/AutomaticConfigDescExample.scala
```diff
@@ -0,0 +1,54 @@
...
+import zio.console.Console.Live.console._
+
+final case class Token(value: String) extends AnyVal
+final case class Aws(region: String, token: Token)
+final case class MyConfig(
+  aws: Aws,
+  port: Int,
+  dburl: String,
+  port2: Option[Double],
+  price: Either[Double, String],
```
jdegoes: It would be good to have not just `Either` here, but a custom sum type, so you can test that sum types are correctly translated into the nested eithers and then iso-mapped to and from the sum type.

jdegoes:
I'm not an expert in Magnolia, but looks good to me. I think we do need derivation for `sealed trait` (sumtypes) in a config. Maybe @ghostdogpr could take a look??

build.sbt
```diff
@@ -70,7 +70,22 @@ lazy val zioConfig =
     )
 
 lazy val examples = module("examples")
+  .settings(
+    skip in publish := true,
+    moduleName := "zio-config-examples",
+    libraryDependencies ++= Seq(
+      "com.propensive" %% "magnolia" % "0.12.0"
```
ghostdogpr: I recommend 0.12.2, it was released just a few days ago and fixes some annoying cases where you get compiler errors a few thousand lines long 😄

zio-config-magnolia/src/main/scala/zio/config/magnolia/ConfigDescriptorProvider.scala
```diff
@@ -0,0 +1,101 @@
...
+          }
+
+        collectAll(loop(caseClass.parameters)).xmap[T](
+          caseClass.rawConstruct
+        )(v => caseClass.parameters.map(_.dereference(v): Any).toList)
+
+      }
+    }
+
+  def dispatch[T](sealedTrait: SealedTrait[ConfigDescriptorProvider, T]): ConfigDescriptorProvider[T] = ???
```
ghostdogpr: As @jdegoes mentioned you need to implement this, otherwise anyone with a sealed trait in their ADT will get no compile error but a runtime exception.

zio-config-magnolia/src/main/scala/zio/config/magnolia/ConfigDescriptorProvider.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  implicit val doubleDescriptor: ConfigDescriptorProvider[Double] =
+    instance(double)
+
+  implicit val intDescriptor: ConfigDescriptorProvider[Int] =
+    instance(int)
+
+  implicit def opt[A: ConfigDescriptorProvider]: ConfigDescriptorProvider[Option[A]] =
+    ConfigDescriptorProvider[A].getDescription(_).optional
+
```
ghostdogpr: You probably planned to add it later, but things like `List` also need a custom instance otherwise it's going to derive it (`List` being a sealed abstract class) and give you weird results.

build.sbt
afsalthaj: Cool. I wil upgrade.

examples/src/main/scala/zio/config/examples/AutomaticConfigDescExample.scala
afsalthaj: Yes, I am in the process of doing it.

zio-config-magnolia/src/main/scala/zio/config/magnolia/ConfigDescriptorProvider.scala
afsalthaj: Yes, I need to include the whole list.

zio-config-magnolia/src/main/scala/zio/config/magnolia/ConfigDescriptorProvider.scala
afsalthaj: Yea, I bumped into it as I was working on. Thanks @ghostdogpr . I can put them soon in this PR.

afsalthaj:
@jdegoes @ghostdogpr 
I have included the sealed trait support as well.
Given here is the example:
https://github.com/zio/zio-config/blob/magnolia/examples/src/main/scala/zio/config/examples/AutomaticConfigDescExample.scala

ghostdogpr:
Nice! You might want to add instances for `List` and `Map` since they are quite common in configuration data.

afsalthaj:
@ghostdogpr One issue with list is what's the key for a list.
As of now the key that we have used is the label of parameter in case class.
We sort of support list configuration in this example, but I am not fully sure how does it map it to an implementation that supports automatic derivation.
https://github.com/zio/zio-config/blob/master/examples/src/main/scala/zio/config/examples/SequenceExample.scala

Anyways, I am still working on it.

afsalthaj:
@mijicd  How do I stop zio-config-magnolia to not compile for 2.11?
I did specifically mention in the build.sbt, that it should be `2.12 and 2.13` only.

mijicd:
@afsalthaj that might not be as easy as it sounds. I'll give it some time tonight though. 

P.S. the simplest way possible is to drop 2.11 completely :).

afsalthaj:
@mijicd Sure, I thought so too. I am disabling 211 for the time being.

afsalthaj:
It seems implementation of list will take time. There will be a follow up PR with the issue. https://github.com/zio/zio-config/issues/146

Issue #86, opened 2019-10-13
Add property tests for nested config


afsalthaj:
Take a look at NestedConfig example, and bring more property laws for it.

bblagojevic94:
@afsalthaj 
I could pick this one.

PR #128, opened 2019-12-02
Add property tests for nested config

bblagojevic94:
This PR is part of the issue #86.

### DIFF
--- a/zio-config/src/test/scala/zio/config/NestedConfigTest.scala
+++ b/zio-config/src/test/scala/zio/config/NestedConfigTest.scala
@@ -1,6 +1,6 @@
 package zio.config
 
-import zio.config.ConfigDescriptor.{ int, nested, string }
+import zio.config.ConfigDescriptor.{ double, int, nested, string }
 import zio.config.NestedConfigTestUtils._
 import zio.config.PropertyTree.{ Leaf, Record }
 import zio.config.helpers._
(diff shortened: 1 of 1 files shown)


afsalthaj:
The tests are looking awesome. I will take a deeper look (to see if we have any missing cases, such as nested config that are optional, either etc).

bblagojevic94:
Thank you, @afsalthaj. Feel free to send me all these cases. I will provide tests to cover all of them.

zio-config/src/test/scala/zio/config/NestedConfigTest.scala
```diff
@@ -0,0 +1,145 @@
...
+          value.db.credentials.fold(Seq.empty[(String, String)]) { c =>
+            val credentialsKey = s"${keys.database}.${keys.credentials}"
+            Seq(
+              s"$credentialsKey.${keys.user}"     -> c.user,
+              s"$credentialsKey.${keys.password}" -> c.password
+            )
+          },
+          Seq(keys.pricing -> value.pricing.toString)
+        ).flatten.toMap
+      )
```
afsalthaj: Did we really need to start with a Seq and then call `toMap` instead of writing a Map explicitly?

afsalthaj: If we know statically what the map is probably manually write a Map. That will make the test case much more easier to reason about.

afsalthaj: Ah! I see why now.

afsalthaj:
I think it is reasonable to merge this now.

zio-config/src/test/scala/zio/config/NestedConfigTest.scala
```diff
@@ -0,0 +1,92 @@
...
+      kDb          <- genKey.filter(_ != kPort)
+      kDbUrl       <- genKey
+      vDbUrl       <- genNonEmptyString(50).map(DbUrl)
+      kCredentials <- genKey.filter(_ != kDbUrl)
+      kUser        <- genKey
+      vUser        <- genNonEmptyString(50)
+      kPassword    <- genKey.filter(_ != kUser)
+      vPassword    <- genNonEmptyString(50)
+    } yield TestParams(kPort, vPort, kDb, kDbUrl, vDbUrl, kCredentials, kUser, vUser, kPassword, vPassword)
+}
```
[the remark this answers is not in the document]
bblagojevic94: I agree I should add other test cases to cover `generateDocs` and `generateDocsWithvalue`. Also, I should try to cover other types of sources such as Hocon notation.

zio-config/src/test/scala/zio/config/NestedConfigTest.scala
```diff
@@ -0,0 +1,145 @@
...
+      )
+  }
+
+  private val genKey = genSymbol(1, 20)
+
+  val genConfigKeys: Gen[Random, KeyParams] =
+    for {
+      kHost        <- genKey
+      kPort        <- genKey.filter(_ != kHost)
+      kUser        <- genKey
```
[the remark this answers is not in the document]
bblagojevic94: `kHost` and `kUser` are not on the same "level" in configuration so they should have the same value?
Also, I was thinking that in this case key's generation may not have any benefit, but it can make code more complicated. Maybe we should have fixed the key's values.

PR #164, opened 2019-12-31
Implement Java property file as a source

afsalthaj:


core/src/main/scala/zio/config/Config.scala
```diff
@@ -36,6 +40,28 @@ object Config {
   ): IO[ReadErrors[Vector[String], String], Config[A]] =
     make[String, String, A](ConfigSource.fromMap(map), configDescriptor)
 
+  // If reading a file, this can have read errors as well as throwable when trying to read the file
+  def fromPropertyFile[A](
```
jdegoes: You probably want to do this on the `Blocking` thread pool to be safe.

core/src/main/scala/zio/config/Config.scala
```diff
@@ -36,6 +40,28 @@ object Config {
...
+    filePath: String,
+    configDescriptor: ConfigDescriptor[String, String, A]
+  ): ZIO[Any, Throwable, Config[A]] =
+    ZIO
+      .bracket(ZIO.effect(new FileInputStream(new File(filePath))))(r => ZIO.effectTotal(r.close()))(inputStream => {
+        ZIO.effect {
```
jdegoes: Consider `ZIO(r.close()).ignore`.

Issue #169, opened 2020-01-03
Incorporate round trip tests for all inbuilt property types


afsalthaj:
We need a proof that we are writing what we read. The foundation of all `writes` in zio-config is PropertyType. Hence we need tests for all the types that we define within property type.

```scala
val variable = "1000"
LongPropertyType.write(LongPropertyType.read(variable)) ==  variable
```

Examples for which the types can go wrong is "duration"
```scala
DurationProperty.write(DurationPropertyType.read("3 day")) is "3 days"
```
I will consider this issue as one of the higher priority one, and is a very good first issue.

zach-albia:
Working on it.

zach-albia:
@afsalthaj what behavior are we expecting `PropertyType`s to have? For instance, should a roundtrip of `"3 day"` yield `"3 day"` instead of `"3 days"`? Is `"3 days"` really wrong? I've also discovered similar cases where a `BigInt` string like `"005"` would roundtrip into `"5"`. There's also the issue of the lack of precision when doing roundtrips on floating point values, though, I guess clients of `PropertyType` instances should know which instance would be best for some use case.

If we really want to preserve the original, wouldn't we have to store the original in some data type?

Personally, I think the current behavior is fine but I want to know your input on this.

afsalthaj:
Probably, in these scenarios, (where a difference is actually ok), we
should read the written value again to see if there is any difference from
the initially read value.

For certain types, even this property won’t be a success, and in such
situations, the only option is to skip those types.

zach-albia:
> Probably, in these scenarios, (where a difference is actually ok), we
should read the written value again to see if there is any difference from
the initially read value.

Here's the generic assertion I wrote to this effect:

`assert(roundTrip(propType, s).map(parse), isRight(equalTo(parse(s))))`, where `s: String` and `parse: String => A`.

> For certain types, even this property won’t be a success, and in such
situations, the only option is to skip those types.

Understood, I'll see how far it goes.

afsalthaj:
@zach-albia Nice one

PR #172, opened 2020-01-04
Add PropertyType tests

zach-albia:
#169

The core idea is to test that string values read, then written by a `PropertyType[String, A]` instance and finally parsed to an `A`, yield the same result as parsing to an `A`. That is, reading from and writing to string values doesn't change the value of the properties they parse to.

Each `PropertyType` instance, except `StringType`, has a valid and invalid string property test.

Generators summary:
- String generators for all numeric types are padded with (0..*) leading zeros for added coverage.
- All case permutations of the strings `"true"` and `"false"` are used to test `BooleanType`.
- `BigInt`s and `BigDecimal`s are generated using numeric strings.....is this okay?
- I based the `BigDecimal` generator loosely on [this diagram](https://i.stack.imgur.com/wmFqa.gif).
- I just reused `genDuration` from helpers.scala.
- Generated URI strings are generic, i.e. loosely based on the [spec here](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Generic_syntax).
- I make heavy use of a `genAppend`, which stitches `Gen[R, String]` values together, which I think should be written in terms of a `Gen.sequence`, as per ZIO issue [#2573](https://github.com/zio/zio/issues/2573).


zach-albia:
Whoops, forgot to test on 2.12.x. Also, the genUri is buggy. Fixing...

core/src/test/scala/zio/config/PropertyTypeTest.scala
```diff
@@ -0,0 +1,269 @@
...
+          propType = DoubleType,
+          genValid = genPadLeadingZeros(Gen.double(Double.MinValue, Double.MaxValue)),
+          parse = _.toDouble
+        ),
+        propertyTypeRoundtripSuite(
+          typeInfo = "BigDecimal",
+          propType = BigDecimalType,
+          genValid = genValidBigDecimalString,
+          parse = BigDecimal(_)
+        ) @@ flaky, // because nums generated could be too large
```
jdegoes: You could solve this by generating only those in a certain range.

core/src/test/scala/zio/config/PropertyTypeTest.scala
```diff
@@ -0,0 +1,271 @@
...
+
+  private val genDigit0To9: Gen[Random, String] = char('0', '9').map(_.toString)
+  private val genDigit1To9: Gen[Random, String] = char('1', '9').map(_.toString)
+
+  /**
+   * Generates a string of up to n digits. Default maxLength digits is 100.
+   */
+  private def genDigits(maxLength: Int = 100): Gen[Random with Sized, String] =
+    sized(n => listOfN(math.min(n, maxLength))(genDigit0To9).map(_.mkString))
+
```
zach-albia: @jdegoes Limiting digits to 100 by default. I went and checked the max and min numbers for this and both `BigInt` and `BigDecimal` and both are fine with them.

afsalthaj:
@jdegoes Are we good to merge this one?

PR #176, opened 2020-01-07
Add scastie link and support for list of primitives in typesafe

afsalthaj:


core/src/main/scala/zio/config/Config.scala
```diff
@@ -53,7 +53,7 @@ object Config {
   def fromPropertyFile[A](
     filePath: String,
     configDescriptor: ConfigDescriptor[String, String, A]
-  ): ZIO[Any, Throwable, Config[A]] =
+  ): Task[Config[A]] =
```
jdegoes: Technically this should be run on the blocking thread pool.

magnolia/src/test/scala/zio/config/magnolia/AutomaticConfigDescriptorSpec.scala
```diff
@@ -120,4 +120,10 @@ object AutomaticConfigTestUtils {
       case (None, Some(v2))     => partialMyConfig + (("anotherDefault", v2.toString))
       case (None, None)         => partialMyConfig
     }
+
+  def genAlpha: Gen[Random, String] =
+    for {
+      n <- Gen.int(1, 10)
+      s <- Gen.listOfN(n)(Gen.char(65, 122))
```
jdegoes: There may be alphaNumeric Gen already

afsalthaj: Yea, Unfortunately it emitted 03 and gave and got parsed as an `Int`.

typesafe/src/main/scala/zio/config/typesafe/TypeSafeConfigSource.scala
```diff
@@ -114,15 +113,37 @@ object TypeSafeConfigSource {
...
+                                          Right(::(h.unwrapped().toString, t.map(_.unwrapped().toString)))
+
+                                        case _ => Left(singleton(ReadError.missingValue[Vector[String], String](path)))
+
+                                      }
+                                    }
+                                }
                         } yield res
                       } else {
                         if (parentConfig.getValue(head).valueType() == ConfigValueType.OBJECT) {
```
afsalthaj: I will need to revisit, however, there is just about lots of `Object` operations in type safe library. So any checks here probably make it much more robust.

typesafe/src/main/scala/zio/config/typesafe/TypeSafeConfigSource.scala
```diff
@@ -114,15 +113,37 @@ object TypeSafeConfigSource {
...
+                                        case h :: t
+                                            if (::(h, t).forall(
+                                              t =>
+                                                t.valueType() != ConfigValueType.NULL ||
+                                                  t.valueType() != ConfigValueType.LIST ||
+                                                  t.valueType() != ConfigValueType.OBJECT
+                                            )) =>
+                                          Right(::(h.unwrapped().toString, t.map(_.unwrapped().toString)))
+
+                                        case _ => Left(singleton(ReadError.missingValue[Vector[String], String](path)))
```
afsalthaj: if user provides things like

```
users = [jon, 1, {hi = bi }]
```
it actually doesn't make sense. But the error should be fatal error providing a custom message. And not missing value. My bad.

afsalthaj:
@jdegoes Could you approve again for this one?

PR #183, opened 2020-01-11
Remove redundant UIO wrappers around returned configuration etc

leigh-perry:
Updating the pattern of instantiating program startup-time shared objects (like configuration and the hypothetical Spark session in the program example)

### DIFF
--- a/examples/src/main/scala/zio/config/examples/RefinedReadConfig.scala
+++ b/examples/src/main/scala/zio/config/examples/RefinedReadConfig.scala
@@ -55,7 +55,7 @@ object RefinedReadConfig extends App {
       val outcome: ZIO[Any, ReadErrors[Vector[String], String], Refined[List[Long], Size[Greater[W.`2`.T]]]] =
         for {
           count  <- Config.fromMap(configMap, nonNegative(int("COUNT")))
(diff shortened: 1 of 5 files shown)


afsalthaj:
Good one!

examples/src/main/scala/zio/config/examples/RefinedReadConfig.scala
```diff
@@ -55,7 +55,7 @@ object RefinedReadConfig extends App {
       val outcome: ZIO[Any, ReadErrors[Vector[String], String], Refined[List[Long], Size[Greater[W.`2`.T]]]] =
         for {
           count  <- Config.fromMap(configMap, nonNegative(int("COUNT")))
-          n      <- count.config.config
+          n      = count.config.config
```
afsalthaj:
```
val source =  ConfigSource.fromMap(configMap)
count <- read(nonNegative(int("COUNT")) from source)

```

afsalthaj: val source =  ConfigSource.fromMap(configMap)
count <- read(nonNegative(int("COUNT")) from source)

PR #187, opened 2020-01-12
Restructured examples / Added Examples to Coproduct.

pme123:
On the proposal to add examples by Afsal Thaj - see https://stackoverflow.com/questions/59670366/how-to-handle-an-adt-sealed-trait-with-zio-config

I added two other CoproductExamples:

* manual
* typesafe

I only made examples for the 2. Version, as the first one did not appeal to me;).

That there was no name clash and as there are now quite a lot examples, I introduced the structure analog to the submodules.

I saw that IntelliJ restructured the imports a bit, I hope this is not a problem.

### DIFF
--- a/examples/src/main/scala/zio/config/examples/CoproductExample.scala
+++ b/examples/src/main/scala/zio/config/examples/CoproductExample.scala
@@ -2,7 +2,7 @@ package zio.config.examples
 
 import zio.DefaultRuntime
 import zio.config.ConfigDescriptor._
-import zio.config.{read, _}
+import zio.config.{ read, _ }
 
 // see Stackoverflow: https://stackoverflow.com/questions/59670366/how-to-handle-an-adt-sealed-trait-with-zio-config
 object CoproductExample extends App {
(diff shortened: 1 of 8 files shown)


afsalthaj:
@pme123 Good one at this stage of zio-config. Thanks for taking the initiative.

examples/src/main/scala/zio/config/examples/CoproductExample.scala
```diff
@@ -15,7 +16,43 @@ object CoproductExample extends App {
   final case class Person(name: String, age: Option[Int])
   final case class Height(height: Long)
 
-  val danceConfig = description[Dance]
+  val personConfig =
+    (string("name") |@| int("age").optional)(Person.apply, Person.unapply)
+
+  val heightConfig =
```
jdegoes: This is not the optimal way to do sum types.

First you should `orElseEither` `<+>` all of the terms together, and then xmap the whole thing. Then you have compile-time help if you add a new subtype—you'll be forced to deal with that. 

With this solution, if a new subtype is added, it will just not work at runtime, you won't get any runtime errors.

pme123: @afsalthaj   gave me two possible solutions: see https://stackoverflow.com/questions/59670366/how-to-handle-an-adt-sealed-trait-with-zio-config/59690405#59690405

So you would suggest that Option 1 is the better one? Or do you have another one?

Issue #193, opened 2020-01-18
Bug when fetching list of products when the product has optional values in it


afsalthaj:
This is applicable in `zio-config-typesafe`
There is an issue when fetching list of case classes, such that if one of the field in the case class is None, reading the config gives wrong results. See the below example

```scala
final case class Aws(region: Option[String])
val desc = description[Aws]

object Main extends App {
  val hocconSourceList =
    hocon(
      Right(
        """
        mylist = [{
         region: abc
        },
        {
         region: null
        }]
       """
      )
    )

  val runtime = new DefaultRuntime {}
     
   runtime.unsafeRun(
     read(nested("mylist")(list(Aws.desc)) from hocconSourceList).either
   )

// Right(List(Aws(None)) (expected: Right(List(Aws("abc")))
}


```

afsalthaj:
This is mainly due to the fact that typesafe config implementation fail early when trying to fetch a list of values after identifying the valueType is a list. It fails, when it hits a `null` which wasn't handled. 

This indicate that fetching a value for a key need to return ::[Option[A]] and not necessarily ::[A]. The option is required to indicate that one of them is `null` (In the above example region is  `null`, and we can't discard this.). In other words, we still need to continue without failing early and still need to return non-failed A in ZIO[E, A]. 

I couldn't get much time investigating this, however, it may need some thinking towards implementing ReadFunctions.

afsalthaj:
An initial version is here:
https://github.com/zio/zio-config/blob/3876998d79731955bf01fce783a13ee30acf25c5/examples/src/main/scala/zio/config/examples/typesafe/NullAndOptionalConfig.scala

afsalthaj:
#197

afsalthaj:
Fixed in #236

PR #197, opened 2020-01-27
Handle null and options in type safe config

afsalthaj:
Solves #193 

With this,  zio-config returns you the positions of all the missing values of port (example below), and not fail-fast with ConfigException that was propagated from typesafe-config.

```scala
case class Db(url: String, port: Int)
case class Details (list: List[Db])

db = [
    {
     url: aa
     port: 12 
    }
    {
     url : ddd
     port: null
    }
    {
     url: dddd
    }
]

```

TODOs
Move examples to property tests
Have configsources test (System properties, System env, java properties etc)


core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -152,9 +152,23 @@ object ConfigSource {
...
+        ZIO
+          .fromOption(
+            map
+              .get(key)
+              .map(
+                r =>
+                  ConfigValue(
+                    ::(
+                      f(r).map(Some(_)).head,
+                      f(r).map(Some(_)).tail
```
afsalthaj: val r =f(r).map...
(r.head...)

core/src/main/scala/zio/config/ReadFunctions.scala
```diff
@@ -11,29 +11,51 @@ private[config] trait ReadFunctions {
...
+                  Right(
+                    ReadFunctions.MissingValuesInList[K, V1, B](list, paths)
+                  ): Either[ReadErrorsVector[K, V1], ReadFunctions.MissingValuesInList[K, V1, B]]
+                )
+              else
+                ZIO.succeed({
+                  val sList = list.flatMap(_.toList)
+                  ::(sList.head, sList.tail)
+                })
+          )
```
afsalthaj: Store indices and number of missing value as part of ReadError (in follow up PR)

afsalthaj: This will be a huge win when dealing with type safe config's list values having empty values in between

afsalthaj:
scala 2.12 is failing to compile with the change, but should be a minor one

afsalthaj:
More changes coming into the code base

Issue #174, opened 2020-01-07
Bug in flattenString() when there is multiple values for same key


afsalthaj:
Ensure the order
Ensure all elements are captured during write

afsalthaj:
Fixed in #236

Issue #168, opened 2020-01-02
Find a better type of Config Errors instead of List 


afsalthaj:
Currently ConfigErrors return a `::(ReadErrors)`. This will not address coproducts (the situations of `OrElseEither`).

Probably we need something along these lines (pseudo code) while emitting the error message

```scala
// Given, this config
sealed trait MyConfig
case class Config1(key1: String) extends MyConfig
case class Config2(key2: Long) extends MyConfig
case class Config3(key3: String, key4: String) extends MyConfig

// After reading, assuming we have a parse error for key2, then we emit the following

SolveAnyOfTheFollowing(
  ::(MissingValue("key1")), 
  ::(ParseError(key = "key2", provided = "thousand",  message = "supposed to be a long")),
  ::(MissingValue("key3"),  MissingValue("key4"))
) 
```

jdegoes:
You want to keep track of the distinction between "all of these need to succeed" and "one of these need to succeed"? If so I can show you the structure that can do that.

afsalthaj:
@jdegoes yes. That’s what I am looking for.

jdegoes:
```scala
sealed trait ErrorTree[+E] {
  def first: E = ???
  // needs custom equals/hashCode!!!
}
object ErrorTree {
  final case class One[E](value: E) extends ErrorTree[E]
  final case class Both[E](left: ErrorTree[E], right: ErrorTree[E]) extends ErrorTree[E]
  final case class OneOf[E](left: ErrorTree[E], right: ErrorTree[E]) extends ErrorTree[E]
}

afsalthaj:
@jdegoes The only concern is we then have a list being represented as nested `Both`.
Could we simplify `Both[E]` to be just `ListOfError[E]` and each error can be `OneOf` or `One`

afsalthaj:
Tracking this in fix_list_bug branch
https://github.com/zio/zio-config/tree/fix_list_bug

It is currently very unpleasant code, and not in a state to raise PR. However, keeping a note here to avoid duplicate work.

afsalthaj:
Fixed in #236

Issue #127, opened 2019-12-01
Make optional config, fail for parse errors instead of returning None


afsalthaj:
Example:

```scala
object ConfigSpec
    extends DefaultRunnableSpec(
      suite("config suite")(
        testM("optional Int example") {

          case class ConfigInternal private (
            domain: String,
            maybePort: Option[Int]
          )
          val envVars: Map[String, String] = Map(
            "DOMAIN" -> "dom1",
            "PORT"   -> "NOT_AN_INT"
          )
          val configDescriptor: ConfigDescriptor[String, String, ConfigInternal] =
            (string("DOMAIN") |@| int(
              "PORT"
            ).optional)(ConfigInternal.apply, ConfigInternal.unapply)

          val configM: IO[ReadErrors[Vector[String], String], config.Config[ConfigInternal]] =
            zio.config.Config.fromMap(envVars, configDescriptor)

          for {
            c        <- configM
            internal <- c.config.config
          } yield assert(internal, equalTo(ConfigInternal("dom1", None)))
          // succeeds but we would expect Optional[Int] field to fail when given data "NOT_AN_INT"
        }
      )
    )
```

drmarjanovic:
@afsalthaj, I would like to take this issue if you agree?

afsalthaj:
@drmarjanovic  I think I missed out your message. Sorry for late response.
Ofcourse, please take it up .Thanks.

afsalthaj:
Feel free to ping me in discord or here anytime.

afsalthaj:
This is now being tracked here, with errors being typed in a better way
https://github.com/zio/zio-config/blob/fix_list_bug/core/src/main/scala/zio/config/ReadFunctions.scala#L115

afsalthaj:
A few behaviours such as;
Optional: 
If there is a parse or unknown error occurred for a key, then its `optional` node will still return error.

OrElseEither: Either[A, B]
If there is any error, parse error or missing value when trying to get, then OrElseEither node tries if it can produce B. For unknown errors, it will avoid trying to get B

`Example: string("key1").orElseEither(int("key2"))`

OrElse: A (it is in fact, Either[A, A])
OrElse node works, if the first key is missing. All the other errors will result in not trying the second key.

`Ex: string("key1").orElse(string("key2)`

afsalthaj:
Fixed in #236

PR #199, opened 2020-02-02
WIP: A whole new bunch of features + bug fixes related to typesafe integration

### LINKED ISSUE #205: Index missing value errors
As of now, the missing values are not indexed. 

A key can exist in multiple places within a config, and user might be expecting values to exist in all these positions. This can happen regardless of the config source, although, it is more relevant to hocon files. 

```

accounts = [
      {
          accountId: jon
      }
      {
          region : us-west
          accountId: chris
      }
      {
          accountId: null
      }
    ]
```

regions are missing from two positions, and hence the errors should be:

```
     AndErrors(
              MissingValue(Vector("accounts", "region"), Some(0)),
              MissingValue(Vector("accounts", "region"), Some(2))
            )

```

afsalthaj:
Fixes #205  #204 #193 #168 #174 #127 

Following are some of the major issues that I am aiming to fix with this PR:

1. Early Errors instead of accumulation when integrated with typesafe-config.
2. Nulls and Optional Values in hocon source was not properly handled. It resulted in errors and `ConfigException` propagated from type-safe.
3. optional parameters resulting in `None` when they encounter parsing errors. 
4. Positions of errors were missing there are missing values in a list of configs. This is mainly applicable for hocon.
5. No support for map yet. PR intends to solve it as well
6. Add more property tests 
7. A single sbt task to run all examples
8. Better Errors - AndErrors and OrErrors

There are a few ugly coding done in this PR. And some necessary changes which I would like to make it as a follow up PR. (Getting rid of some intermediate structures and better error structures)

This will be fixed as we move on. In this PR, we are aiming at a few major functionalities and bug fixes.

PS: Build issues are in 2.12, and it doesn't exist in 2.13. I will fix this! [EDIT: FIXED]


core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -4,15 +4,39 @@ import zio.{ IO, ZIO }
...
+ * ConfigSource represents the source to the config-description, which can be used for reading the values.
+ *
+ * @param getConfigValue : For a key, ConfigValue may or may not exist.
+ *                       Or in the process of fetching it, the computation can fail with some unknown error
+ *
+ * @param sourceDescription: Every ConfigSource is associated with a description
+ */
 final case class ConfigSource[K, V](
-  getConfigValue: Vector[K] => IO[ReadErrorsVector[K, V], ConfigValue[V]],
+  getConfigValue: Vector[K] => IO[ReadError.Unknown[Vector[K]], Option[ConfigValue[V]]],
```
jdegoes:
```suggestion
  getConfigValue: Vector[K] => Task[Option[ConfigValue[V]]],
```

In the code that calls `getConfigValue`, pack it into `ReadError.Unknown[Vector[K]]`.

jdegoes: Consider:

```scala
IO[Option[Throwable], ConfigValue[V]]
```

But only if treating `None` as `Throwable` is common.

core/src/main/scala/zio/config/ReadError.scala
```diff
@@ -5,16 +5,40 @@ sealed trait ReadError[+K, +V] {
 }
 
 object ReadError {
-  final case class MissingValue[K](key: K)                                extends ReadError[K, Nothing]
+  final case class MissingValue[K](key: K, positions: Option[Int] = None) extends ReadError[K, Nothing]
```
jdegoes:
```suggestion
  final case class MissingValue[K](key: K, position: Option[Int] = None) extends ReadError[K, Nothing]
```

afsalthaj:
I think this PR gels well with adding the most important feature, the support for maps. We are anyways changing the structure of `ConfigSource`. It's the right time to fit this in.

afsalthaj:
@jdegoes In supporting maps, could the `ConfigSource` be returning `PropertyTree`?
Or is that too much of a complexity for users defining their own `ConfigSource`?

jdegoes:
>  In supporting maps, what do you think ConfigSource be returning PropertyTree?
Or is that too much of a complexity for users defining their own ConfigSource?

Sure, it could. You can try it and see where it leads. I think it will work.

build.sbt
```diff
@@ -84,14 +84,31 @@ lazy val zioConfigRefined =
     )
     .dependsOn(zioConfig % "compile->compile;test->test")
 
+lazy val runAllExamples = taskKey[Unit]("Run all main classes in examples module")
```
afsalthaj: An sbt task to quickly run all examples

core/src/main/scala/zio/config/Config.scala
```diff
@@ -33,21 +33,21 @@ object Config {
...
-  ): IO[ReadErrors[Vector[String], String], Config[A]] =
+  ): IO[ReadErrors[Vector[String]], Config[A]] =
     make(ConfigSource.fromEnv(valueDelimiter), configDescriptor)
 
   def fromMap[A](
     map: Map[String, String],
     configDescriptor: ConfigDescriptor[String, String, A],
     pathDelimiter: String = "."
-  ): IO[ReadErrors[Vector[String], String], Config[A]] =
+  ): IO[ReadErrors[Vector[String]], Config[A]] =
```
afsalthaj: ReadErrors knew about `V` only for message purpose. I don't think its worth it considering the complexity that it caused across the entire applications.

ReadErrors is associated with a key.  The only time value is used is to print out a message of parse errors

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -4,15 +4,21 @@ import zio.{ IO, ZIO }
 import zio.system.System.Live.system
 import java.{ util => ju }
 
-final case class ConfigValue[A](value: ::[A])
+final case class ConfigValue[A](value: ::[Option[::[A]]])
```
afsalthaj: Lists can exist 

[
   {
   key1: [ value1, value2],
   key2: [value2, value3]
  },
   {
   key1: [ value3, value4],
   key2: [value5, value6]
  },
]

This opens the space for dynamic map

example

```
KAFKA_BOOTSTRAP_SERVERS = "x"
KAFKA_CG_NAME = "y"
```

and we could do

```
nested("kafka")(map)
```

to obtain

```
Map(
   BOOTSTRAP_SERVERS -> x,
   CG_NAME -> y
 )
```

core/src/main/scala/zio/config/ConfigSource.scala
afsalthaj: Will do this

core/src/main/scala/zio/config/ReadError.scala
```diff
@@ -1,20 +1,11 @@
...
-  def parseError[K, V](key: K, provided: V, message: String): ReadError[K, V] =
-    ParseError(key, provided, message)
-
-  def fatalError[K, V](key: K, cause: Throwable): ReadError[K, V] =
-    FatalError(key, cause)
+  final case class MissingValue[K](key: K, positions: Option[Int] = None)           extends ReadError[K]
+  final case class ParseError[K, V](key: K, message: String)                        extends ReadError[K]
+  final case class Unknown[K](key: K, cause: Throwable)                             extends ReadError[K]
+  final case class OrErrors[K](leftErrors: ReadError[K], rightErrors: ReadError[K]) extends ReadError[K]
+  final case class AndErrors[K](errors: ::[ReadError[K]])                           extends ReadError[K]
```
afsalthaj: There should be one more called `ListErrors` which is a list of errors dealing with transforming values of a list.
`AndErrors` should then be just `left` and `right` similar to `OrErrors`

core/src/main/scala/zio/config/ReadFunctions.scala
```diff
@@ -2,129 +2,260 @@ package zio.config
...
-                .bimap(
-                  err =>
-                    singleton(
-                      ReadError.fatalError[Vector[K], V1](paths, new RuntimeException(err))
-                    ),
-                  res => res
+                .succeed(
+                  a.flatMap(
+                    values =>
+                      seqEitherCons(mapCons(values)(f)) match {
```
afsalthaj: This can be improved actually. Collect the errors here as well.

In that way 

   [
       {  
          k : [v1, v2, v3]
       }   
    
       
       {  
          k : null
       }   

   ]
The error when parsing each element in first k, and the error when trying to fetch the second k should all be accumulated

core/src/main/scala/zio/config/ReadError.scala
```diff
@@ -1,20 +1,11 @@
 package zio.config
 
-sealed trait ReadError[+K, +V] {
-  def key: K
-}
+sealed trait ReadError[+K]
```
afsalthaj: Implement `def prettyPrint`

afsalthaj: Probably, we could reuse the logic of pretty print of `Cause` in ZIO

build.sbt
```diff
@@ -84,14 +84,31 @@ lazy val zioConfigRefined =
...
+
+          val x = c.map(cc => {
+            Def.task {
+              runs.toTask(s" ${cc}").value
+            }
+          })
+
+          Def.sequential(x)
+        })
+        .value
```
afsalthaj: Mainly intended for contributors.
We have 21 examples now, each one of them has an assertion to tell the user what's the output looking like,
After making changes, if someone needs to run all examples with in sbt shell, 

```
sbt runAllTasks
```
If any of the assertions fail, the task stops

core/src/main/scala/zio/config/ConfigSource.scala
afsalthaj: This I think could be

```scala
sealed trait Value[A]

case class Leaf(l: Option[::[A]]) extends Value[A]
case class Segment(x: ::[Value[A]]) extends Value[A]


```

core/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -78,6 +78,22 @@ sealed trait ConfigDescriptor[K, V, A] { self =>
     }
     loop(self)
   }
+
+  def mapKey(f: K => K): ConfigDescriptor[K, V, A] = {
+    def loop[B](config: ConfigDescriptor[K, V, B]): ConfigDescriptor[K, V, B] = config match {
```
jdegoes: Probably worth using `@tailrec` here

core/src/main/scala/zio/config/ConfigSource.scala
jdegoes: I think in the long run, it may be simpler to migrate to `PropertyTree`. We can have helpers for "flat" property formats that make it easier to construct `ConfigSource`.

PR #203, opened 2020-02-13
Update ZIO to 1.0.0-RC18

ioleo:
Same as with [interop-cats#145](https://github.com/zio/interop-cats/pull/145). Been playing around and needed to update `zio-config` to latest ZIO. Tested only on `2.13.1`.

This PR is meant to hang around open until next ZIO release. I'll try to keep it up to date as ZIO Core evolves.

### DIFF
--- a/core/src/main/scala/zio/config/ConfigDescriptor.scala
+++ b/core/src/main/scala/zio/config/ConfigDescriptor.scala
@@ -6,47 +6,6 @@ import java.net.URI
 import zio.config.ConfigDescriptor._
 
 sealed trait ConfigDescriptor[K, V, A] { self =>
-  final def zip[B](that: => ConfigDescriptor[K, V, B]): ConfigDescriptor[K, V, (A, B)] =
-    ConfigDescriptor.Zip(self, that)
-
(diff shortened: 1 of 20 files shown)


core/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -6,60 +6,45 @@ import java.net.URI
 import zio.config.ConfigDescriptor._
```
ioleo: All I did here is to sort methods alphabetically.

core/src/main/scala/zio/config/package.scala
```diff
@@ -1,11 +1,14 @@
 package zio
 
 package object config extends ReadFunctions with WriteFunctions with ConfigDocsFunctions {
+
+  type Config[A] = Has[Config.Service[A]]
```
ioleo: Following the convention from ZIO Core.

afsalthaj: Sure 👍

afsalthaj:
Thanks @ioleo for the PR. Looking forward

ioleo:
Starting to upgrade the PR to `1.0.0-RC18`.

PR #229, opened 2020-03-17
Add succeed constructor to Config.Service

vpavkin:
Seems useful by itself, but my particular case is simplifying manipulation of already existing service instances.

A particular example is to zoom into a root config layer to get a sub-config for a particular component. Currently you'd have to write smthing like:

```scala
val configLayer = TypesafeConfig.fromDefaultLoader(appConfigDesc)
val dbConfigLayer = configLayer.map(c => Has(new Config.Service[DbConfig] { def config = c.get.config.db }) )
val apiConfigLayer = configLayer.map(c => Has(new Config.Service[ApiConfig] { def config = c.get.config.api }) 
```

With `pure` it becomes much simpler:

```scala
val configLayer = TypesafeConfig.fromDefaultLoader(appConfigDesc)
val dbConfigLayer = configLayer.map(c => Has(Config.pure(c.get.config.db)))
val apiConfigLayer = configLayer.map(c => Has(Config.pure(c.get.config.api)))
```

### DIFF
--- a/core/src/main/scala/zio/config/Config.scala
+++ b/core/src/main/scala/zio/config/Config.scala
@@ -12,6 +12,10 @@ object Config {
     def config: A
   }
 
+  def pure[A](a: A): Service[A] = new Service[A] {
+    def config: A = a
+  }
+
   def make[K, V, A](
     source: ConfigSource[K, V],
     configDescriptor: ConfigDescriptor[K, V, A]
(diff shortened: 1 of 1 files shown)


afsalthaj:
Thanks @vpavkin . LGTM.  Let's wait for a review from @jdegoes as well

core/src/main/scala/zio/config/Config.scala
```diff
@@ -12,6 +12,10 @@ object Config {
     def config: A
   }
 
+  def pure[A](a: A): Service[A] = new Service[A] {
```
jdegoes:
```suggestion
  def succeed[A](a: => A): Service[A] = new Service[A] {
```

vpavkin: Renamed, thanks!

PR #235, opened 2020-03-28
Add support for command line arguments-based config, needed for Beam …

leigh-perry:
…programs etc


afsalthaj:
Pretty cool one! This makes it really stand out.

afsalthaj:
@leigh-perry Thanks for the contribution. However, I would request you to wait for 1-2 days ( to merge my upcoming big PR (hopefully)). Is it ok?

leigh-perry:
No problem - I need to fix the cross build test runs anyway :-)

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -60,6 +60,34 @@ object ConfigSource {
       SystemEnvironment :: Nil
     )
 
+  /** Forming configuration from command line arguments, eg `List(--param1=xxxx, --param2=yyyy)` */
+  def fromArgs(args: List[String], separator: Option[String] = None): ConfigSource[String, String] = {
+    val map =
+      args
+        .filter(s => s.startsWith("--") && s.contains("="))
```
leigh-perry: This ignores any args that aren't switches, ie of the form "--param=value".
The idea was to allow command line switches to be processed while ignoring other command args.
An alternative approach would be to treat them as an error.
What do you think?

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -60,6 +60,34 @@ object ConfigSource {
...
+        .map(s => s.substring(2).split('=').toList)
+        .flatMap {
+          _ match {
+            case s1 :: s2 :: Nil => List((s1, s2))
+            case _               => Nil
```
leigh-perry: Similarly this ignores invalidly formed switch-style arguments

core/src/main/scala/zio/config/ConfigSource.scala
jdegoes: I think we should take a different approach here: pack all of the command-line arguments into multiple property lists.

Using `PropertyTree.mergeAll`, we can merge a bunch of command line options into the smallest possible set of property trees, and then use those property trees to perform lookup.

All the hard logic is embedded inside `mergeAll` / `merge`, so if the command-line options form multiple conflicting schemas, then they'll be broken apart into multiple property trees.

Just loop through all command line arguments, looking for the form:

```
--xyz=xyz
--xyz xyz
-xyz=xyz
-xyz xyz
```

Using `.` as the path separator for the keys.

Extract each pair as a separate `PropertyTree`. This will give you a `List`, which you can call `mergeAll` on to "compact" it into compatible schemas.

Probably, you should sort the list by order of decreasing size. This will ensure that in the event of conflicts or unrelated command line options, you get the "biggest" merge.

Then you can construct a `ConfigSource` that chooses between the different `PropertyTree`, starting from the biggest and proceeding to the smallest.

Finally, I'd add a `Config.fromArgs` helper method that does what all the other helper methods do there.

jdegoes:
@leigh-perry 

I think we should take a different approach for the parsing of command-line overrides: pack all of the command-line arguments into multiple property lists.

Using `PropertyTree.mergeAll`, we can merge a bunch of command line options into the smallest possible set of property trees, and then use those property trees to perform lookup.

All the hard logic is embedded inside `mergeAll` / `merge`, so if the command-line options form multiple conflicting schemas, then they'll be broken apart into multiple property trees.

Just loop through all command line arguments, looking for the form:

```
--xyz=xyz
--xyz xyz
-xyz=xyz
-xyz xyz
```

Using `.` as the path separator for the keys.

Extract each pair as a separate `PropertyTree`. This will give you a `List`, which you can call `mergeAll` on to "compact" it into compatible schemas.

Probably, you should sort the list by order of decreasing size. This will ensure that in the event of conflicts or unrelated command line options, you get the "biggest" merge.

Then you can construct a `ConfigSource` that chooses between the different `PropertyTree`, starting from the biggest and proceeding to the smallest.

Finally, I'd add a `Config.fromArgs` helper method that does what all the other helper methods do there.

afsalthaj:
@leigh-perry If needed, we can have follow up PRs for handling multiple cases  (and merging logic using property tree which John mentioned).

afsalthaj:
Once you are ready, we can merge this one.
IMO, since it's a huge-addition as a feature into the zoo-config, let's tag it as experimental in the methods in `Config` and `ConfigSource` apis.

If any scenarios do fail, please comment them as well. In both classes.

core/src/main/scala/zio/config/Config.scala
```diff
@@ -1,10 +1,18 @@
 package zio.config
 
+import java.util.Properties
+
 import zio.system.System
 import zio.{ Layer, Tagged, ZIO, ZLayer }
-import java.util.Properties
 
 object Config {
+  def fromArgs[K, V, A](
```
afsalthaj, line 9: Huge addition, IMO, tag them as experimental

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -33,6 +33,26 @@ object ConfigSource {
...
     ConfigSource(_ => PropertyTree.empty, Nil)
 
+  /**
+   * Forming configuration from command line arguments, eg `List(--param1=xxxx, --param2=yyyy)`
+   *
+   * Pack all of the command-line arguments into multiple property lists. Using PropertyTree.mergeAll, we can merge a
+   * bunch of command line options into the smallest possible set of property trees, and then use those property trees
+   * to perform lookup.
+   */
+  def fromArgs(
```
afsalthaj, line 43: Huge addition, IMO, it's a good idea to tag them as experimental.

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -209,6 +222,26 @@ object ConfigSource {
...
+   * -xyz xyz
+   */
+  private[config] def argsAsKeyValues(args: List[String]): List[(String, String)] =
+    args.flatMap { s =>
+      if (s.startsWith("-") && s.contains("="))
+        s.substring(2).split('=').toList match {
+          case k :: v :: Nil => List(k, v)
+          case _             => Nil
+        }
+      else List(s)
```
afsalthaj, line 240: We expect some sort of smart constructors here.
We don't want to end up having key-value pairs for `-=hello` etc

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -53,16 +73,9 @@ object ConfigSource {
...
   ): ConfigSource[String, String] =
-    fromMapInternal(map)(
-      x => {
-        val listOfValues = valueDelimter.fold(List(x))(delim => x.split(delim).toList)
-        ::(listOfValues.head, listOfValues.tail)
-      },
-      keyDelimiter,
-      source
-    )
+    fromMapInternal(map)(splitDelimited(_, valueDelimiter), keyDelimiter, source)
```
afsalthaj, line 78: I think, it makes sense to push this feature as a map for now. Let's go with incremental approach 👍 
However, we can revisit soon. I need to spend some time investing on PropertyTreeing it directly. I will let you know

core/src/test/scala/zio/config/testsupport/MapConfigTestSupport.scala
```diff
@@ -0,0 +1,110 @@
...
+  object JobConfig {
+    val descriptor: ConfigDescriptor[String, String, JobConfig] =
+      nested("job")((DataflowConfig.descriptor.optional |@| boolean("supervise"))(JobConfig.apply, JobConfig.unapply))
+  }
+
+  def stringN(min: Int, max: Int): Gen[Random with Sized, String] =
+    for {
+      n <- Gen.int(min, max)
+      s <- Gen.stringN(n)(alphaNumericChar)
+    } yield s
```
afsalthaj, line 108: Fantastic tests @leigh-perry .
I remember discussing this. It has got lot better. Thanks

afsalthaj:
A great addition. Thanks @leigh-perry  !!

I have asked for small changes (comments, tagging them as experimental, not sure if Scala has API doc annotation, if you could do a research that will be great)

core/src/main/scala/zio/config/ConfigSource.scala
afsalthaj: Or may be not. Let's see. It's good for now. But could you add an issue page to have a follow up PR on command-line args, to keep the users know we are working on it.

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -1,35 +1,56 @@
...
-  final def orElse(that: => ConfigSource[K, V]): ConfigSource[K, V] =
+  def orElse(that: => ConfigSource[K, V]): ConfigSource[K, V] =
     ConfigSource(
       k => getConfigValue(k).getOrElse(that.getConfigValue(k)),
       if (self.sourceDescription == that.sourceDescription) self.sourceDescription
       else self.sourceDescription ++ that.sourceDescription
     )
 
-  final def <>(that: => ConfigSource[K, V]): ConfigSource[K, V] =
+  def <>(that: => ConfigSource[K, V]): ConfigSource[K, V] =
```
afsalthaj, line 24: All of our methods are final. As a standard

core/src/main/scala/zio/config/ConfigSource.scala
afsalthaj, line 43: Huge addition, IMO, it's a good idea to tag them as experimental.

core/src/main/scala/zio/config/ConfigSource.scala
leigh-perry: `ConfigSource` class is final, so methods are already final.

core/src/main/scala/zio/config/ConfigSource.scala
leigh-perry: I am using `fromMapInternal` as a smart constructor of `ConfigSource` – the map only contains one key-value. The `ConfigSource`s are all then processed using `mergeAll[K, V](sources: Iterable[ConfigSource[K, V]])`.
Is there a better way of constructing a `ConfigSource`?

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -33,6 +33,29 @@ object ConfigSource {
...
     ConfigSource(_ => PropertyTree.empty, Nil)
 
+  /**
+   * Forming configuration from command line arguments, eg `List(--param1=xxxx, --param2=yyyy)`
+   *
+   * Pack all of the command-line arguments into multiple property lists. Using PropertyTree.mergeAll, merge a
+   * bunch of command line options into the smallest possible set of property trees, and then use those
+   * property trees to perform lookup.
+   *
+   * This is an experimental implementation for handling of key/value switches, and is not a
```
afsalthaj, line 43: Could you have this comment in both `fromArgs` method in the project.

Let's say `EXPERIMENTA` in the first line of  the comment

leigh-perry: Done

PR #236, opened 2020-03-29
Change the read and write in terms of PropertyTree

afsalthaj:
TODO:

WriteFunctions.scala `Sequence` node seem to be buggy.
`reduceInner` in PropertyTree of lists are buggy, when it comes to empty list. As of now empty list is tagged as errors to avoid the scenario.


build.sbt
```diff
@@ -84,17 +84,38 @@ lazy val zioConfigRefined =
...
+
+          val x = c.map(cc => {
+            Def.task {
+              runs.toTask(s" ${cc}").value
+            }
+          })
+
+          Def.sequential(x)
+        })
+        .value
```
afsalthaj: All of the examples has assertions to provide the users a view of the return types + validating the examples itself.  For contributors, `runAllExamples` will run all assertions in the examples quickly in parallel

core/src/main/scala/zio/config/ReadError.scala
```diff
@@ -1,20 +1,36 @@
...
+      case ReadError.AndErrors(list)                => ReadError.AndErrors(list.map(_.atKey(key)))
+    }
+  def atIndex(index: Int): ReadError[A] =
+    self match {
+      case ReadError.MissingValue(path)             => ReadError.MissingValue(path :+ Left(index))
+      case ReadError.FormatError(path, message)     => ReadError.FormatError(path :+ Left(index), message)
+      case ReadError.ConversionError(path, message) => ReadError.ConversionError(path :+ Left(index), message)
+      case ReadError.OrErrors(list)                 => ReadError.OrErrors(list.map(_.atIndex(index)))
+      case ReadError.AndErrors(list)                => ReadError.AndErrors(list.map(_.atIndex(index)))
+    }
```
afsalthaj: This is not yet used

core/src/main/scala/zio/config/ReadFunctions.scala
```diff
@@ -1,132 +1,278 @@
...
+                case PropertyTree.Leaf(Left(value)) => PropertyTree.Sequence(List(PropertyTree.Leaf(Left(value))))
+              }
+            )
+
+          sequenceErrors.map(_.map(_ :: Nil)).reduceInner[Either[ReadError[K], List[B]]] {
+            case (Right(l), Right(r)) => Right(l ++ r)
+            case (Left(l), Right(_))  => Left(l)
+            case (Right(_), Left(r))  => Left(r)
+            case (Left(l), Left(r))   => Left(ReadError.AndErrors(l :: r :: Nil))
+          }
```
afsalthaj: Indexing of errors is yet to be done

core/src/main/scala/zio/config/Config.scala
```diff
@@ -4,56 +4,52 @@ import java.io.{ File, FileInputStream }
 import java.util.Properties
 
 import zio.system.System
-import zio.{ system, IO, Tagged, ZEnv, ZIO, ZLayer }
+import zio.{ Layer, Tagged, ZIO, ZLayer }
```
jdegoes: For these methods, I'd make sure that:

1. For every constructor `ConfigSource` (`fromXYZ`), we have a corresponding constructor here, and with the same name.
2. That all methods in this object are alphabetized, with operators appearing first; and private methods moved to a separate section at the end of the object.

core/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -80,6 +80,22 @@ sealed trait ConfigDescriptor[K, V, A] { self =>
   final def xmap[B](to: A => B, from: B => A): ConfigDescriptor[K, V, B] =
     self.xmapEither(a => Right(to(a)), b => Right(from(b)))
 
+  def mapKey(f: K => K): ConfigDescriptor[K, V, A] = {
```
jdegoes: I'd make sure all methods are alphabetical in here, following the preceding conventions:

1. Operators before named methods.
2. Private / protected methods pushed to the very end.

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -1,166 +1,108 @@
...
-                ConfigValue(consOfValues)
-              })
-          )
-          .flatMap(IO.fromOption(_))
-          .mapError[ReadErrorsVector[String, String]](_ => singleton(ReadError.MissingValue(path)))
-          .provideLayer(System.live)
-      },
-      SystemEnvironment :: Nil
-    )
+  def fromSystemEnv(valueSeparator: Option[Char] = None): UIO[ConfigSource[String, String]] =
```
jdegoes: It will lead to a nicer API if we have two versions:

```scala
def fromSystemEnv: ...
def fromSystemEnv(valueSeparator: Char)
```

This way, you don't have to wrap in `Option`, and don't have to use empty parens for the "none" case.

Any method that takes a single parameter with default argument can be factored like this.

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -1,166 +1,108 @@
...
-              })
-          )
-          .flatMap(IO.fromOption(_))
-          .mapError[ReadErrorsVector[String, String]](_ => singleton(ReadError.MissingValue(path)))
-          .provideLayer(System.live)
-      },
-      SystemProperties :: Nil
+  def fromSystemProperties(valueSeparator: Option[Char] = None): UIO[ConfigSource[String, String]] =
+    for {
+      systemProperties <- UIO.effectTotal(java.lang.System.getProperties)
```
jdegoes: We should create tickets in `zio` to expose all system properties and environment variables in maps, so we can just use those methods here in this project, without re-wrapping.

jdegoes: But this is fine for now!

core/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -1,12 +1,98 @@
...
+    case Leaf(value)     => Leaf(f(value))
+    case Record(v)       => Record(v.map { case (k, tree) => (k, tree.map(f)) })
+    case Sequence(value) => Sequence(value.map(_.map(f)))
+    case Empty           => Empty
+  }
+
+  final def getOrElse[K1 >: K, V1 >: V](tree: => PropertyTree[K1, V1]): PropertyTree[K1, V1] =
+    if (self == PropertyTree.empty) tree else self
+
+  def mapEmptyToError[E, V2](f: => E): PropertyTree[K, Either[E, V]] = self match {
```
jdegoes: `final def`

core/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -1,12 +1,98 @@
 package zio.config
 
-import zio.config.PropertyTree.{ Empty, Leaf, Record }
-import zio.config.PropertyTree.Sequence
+import scala.collection.immutable.Nil
+import PropertyTree._
 
 sealed trait PropertyTree[+K, +V] { self =>
+  final def map[V2](f: V => V2): PropertyTree[K, V2] = self match {
```
jdegoes: I'd make sure all these methods are alphabetized.

core/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -1,12 +1,98 @@
...
+    if (self == PropertyTree.empty) tree else self
+
+  def mapEmptyToError[E, V2](f: => E): PropertyTree[K, Either[E, V]] = self match {
+    case Leaf(value)        => Leaf(Right(value))
+    case Record(v)          => Record(v.map { case (k, tree) => (k, tree.mapEmptyToError(f)) })
+    case PropertyTree.Empty => PropertyTree.Leaf(Left(f))
+    case Sequence(value)    => Sequence(value.map(_.mapEmptyToError(f)))
+  }
+
+  final def zipWith[K1 >: K, V2, V3](that: PropertyTree[K1, V2])(f: (V, V2) => V3): PropertyTree[K, V3] =
```
jdegoes: We should test the following laws:

```scala
tree.zipWith(tree)(_._1) == tree
tree.zipWith(tree)(_._2) == tree
```

for all `tree`

afsalthaj: Sure

afsalthaj: Done

core/src/main/scala/zio/config/ConfigDescriptor.scala
afsalthaj: Done

jdegoes:
@afsalthaj `mdoc` is currently failing, but otherwise, looks good to merge! We can deal with other things in a follow-up.

afsalthaj:
@jdegoes Fixed the mdoc. I think an update requires another approval from you

PR #245, opened 2020-04-03
WIP: Write config back to hocon

afsalthaj:
Note: ConfigSource is big change, hence hidden. Please make sure to review


afsalthaj:
@jdegoes 
This is to get your initial suggestion on placement of code.

Mainly the new functionality is writing the config back to Hocon structure. 
The function requires typesafe-config library, because the result of the function is `type safe.config.Config` and its unto the user then to render it as string. For this reason, it is placed in typesafe module in zio-config.

examples/src/main/scala/zio/config/examples/ListExample.scala
```diff
@@ -35,4 +34,40 @@ object ListExample extends App with EitherImpureOps {
...
+//      "regions" : [
+//    # hardcoded value
+//      "australia",
+//    # hardcoded value
+//      "canada",
+//    # hardcoded value
+//      "usa"
+//    ],
+//    # hardcoded value
+//      "xyz" : "something"
```
afsalthaj, line 69: It's sort of baffling to see why that `hardcoded value` came in.
I tried it manually, and still the structure produces extraneous comments.

afsalthaj: However it doesn't affect result (meaning, we are able to read back)

afsalthaj: It comes only when rendering as string. No information about doc/comment is kept in the actual `SimpleConfigObject` structure

typesafe/src/main/scala/zio/config/typesafe/PropertyTreeFunctions.scala
```diff
@@ -0,0 +1,72 @@
...
+  def partitionWith[K, V, A](
+    trees: List[PropertyTree[K, V]]
+  )(pf: PartialFunction[PropertyTree[K, V], A]): List[A] =
+    trees.map {
+      case tree if pf.isDefinedAt(tree) => pf(tree) :: Nil
+      case _                            => Nil
+    }.foldLeft(List.empty[A]) {
+      case (accLeft, left) => (accLeft ++ left)
+    }
+}
```
afsalthaj, line 72: There are multiple entry points to define configs in type safe config. During recursion, we will have to touch all these entry points in different scenarios. I couldn't make it any simpler after studying typesafe-config API

core/src/test/scala/zio/config/PropertyTreeReduceInnerTest.scala
```diff
@@ -59,14 +59,14 @@ object PropertyTreeTest
...
-        },
-        testM("reduceInner should keep the empty nodes as it is") {
-          check(Gen.int(1, 20)) { input =>
-            val listOfEmpty = List.fill(input)(PropertyTree.empty)
-            assert(Sequence(listOfEmpty).map((a: Any) => a :: Nil).reduceInner(_ ++ _))(
-              equalTo(Sequence(listOfEmpty).map((a: Any) => a :: Nil))
-            )
-          }
         }
+        // testM("reduceInner should keep the empty nodes as it is") {
```
afsalthaj, line 63: Started failing, but I am yet to figure out if it’s a false negative

typesafe/src/main/scala/zio/config/typesafe/PropertyTreeFunctions.scala
```diff
@@ -0,0 +1,72 @@
...
+                  .toConfig
+
+            case None => ConfigFactory.empty()
+          }
+      }
+
+    loop(tree, Vector.empty).root()
+  }
+
+  def partitionWith[K, V, A](
```
afsalthaj, line 63: Rename this function

afsalthaj:
@leigh-perry Would you mind approving this PR for the time being. It does have a bug fix which we don't need to hold up.

We can revisit on changes in follow up PRs

Issue #248, opened 2020-04-05
Make ConfigSource.fromArgs pure and not returning UIO


afsalthaj:

afsalthaj:
Fixed

PR #250, opened 2020-04-06
Fix test case in args source

### LINKED ISSUE #246: Accumulate list contents by multiple cmd line args with same key

afsalthaj:
Fixes #246 

Also `--key` is not an empty list. It is a missing value, that nothing is provided for the key `key`.
Hence property test for args is changes from `0 until count` to `1 to count` for generating random lists


core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -51,10 +51,14 @@ object ConfigSource {
     valueDelimiter: Option[Char]
   ): UIO[ConfigSource[String, String]] =
     UIO.effectTotal(argsAsKeyValues(args)).map { kvs =>
-      mergeAll(
-        kvs.map { kv =>
-          fromMapInternal(Map(kv))(splitDelimited(_, valueDelimiter), keyDelimiter, "command line arg")
-        }
+      ConfigSource.fromPropertyTrees(
```
leigh-perry, line 54: https://github.com/zio/zio-config/issues/248
you can get rid of UIO where modifying this method

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -51,10 +51,14 @@ object ConfigSource {
...
+        PropertyTree.mergeAll(
+          kvs.map {
+            case (str, str1) =>
```
leigh-perry, line 57: key, value instead of str, str1
:-)

core/src/test/scala/zio/config/ArgsListAccumulationTest.scala
```diff
@@ -0,0 +1,41 @@
+package zio.config
+
+import zio.ZIO
+import zio.config.ConfigDescriptor.{ int, list }
+import zio.test.Assertion._
+import zio.test.environment.TestEnvironment
+import zio.test.{ DefaultRunnableSpec, _ }
+
+object ArgsListAccumulationTest extends DefaultRunnableSpec {
```
leigh-perry, line 9: My idea for this test was to test all the applicable input media, like properties, args etc.
`ArgsListAccumulationTest` should be `ListAccumulationTest`

core/src/main/scala/zio/config/ConfigSource.scala
afsalthaj: Yea sure

core/src/test/scala/zio/config/ArgsListAccumulationTest.scala
afsalthaj: Not sure. You were using a different set of combinators + logic to form the args source which is unlike properties, hocon.

afsalthaj: If the assumption was, every source is a map, then we have a single test case. But its not the case.

afsalthaj: For example, the above logic for command line args, we are not using Map anymore.
Or to put it this way, every source has their own way of forming the tree.
For example the way we form the tree from hocon isn't the same as we form the tree from a map/sys.env.

Hence we need to test per source.

Ideally, combinators in PropertyTree are proven to be working. But whether the source is making use of the correct combinators lead to testing them

afsalthaj: PropertyTree.mergeAll was missing from ArgsSource. Now that we added, and we test that explicitly.

core/src/main/scala/zio/config/ConfigSource.scala
afsalthaj: auto-complete from intellij. fixed

afsalthaj:
We will have another follow up PR that avoids some restrictions (being able to use only k=v version) in CommandLine

PR #259, opened 2020-04-06
Fix 258, 257 allow ConfigDescriptor without path

senia-psm:
Allow ConfigDescriptor without path
Add ConfigDocs.Sequence

`string(path)` becomes `string.at(path)`
`list(string(path))` -> `list(string).at(path)`

### DIFF
--- a/core/src/main/scala/zio/config/Config.scala
+++ b/core/src/main/scala/zio/config/Config.scala
@@ -34,9 +34,9 @@ object Config {
    *
    *  val credentials = (string("username") |@| string("password"))(Credentials.apply, Credentials.unapply)
    *
-   *  final case class Config(databaseCredentials: Credentials, vaultCredentials: Credentials, regions: List[String, users: List[String])
+   *  final case class Config(databaseCredentials: Credentials, vaultCredentials: Credentials, regions: List[String], users: List[String])
    *
-   *  (nested("db") { credentials } |@| nested("vault") { credentials } |@| list(string("regions") |@| list(string("user"))(Config.apply, Config.unapply)
+   *  (nested("db") { credentials } |@| nested("vault") { credentials } |@| list(string("regions")) |@| list(string("user")))(Config.apply, Config.unapply)
    *
    *  // res0 Config(Credentials(1, hi), Credentials(3, 10), List(111, 122), List(k1, k2))
    *
--- a/core/src/main/scala/zio/config/ConfigDescriptor.scala
+++ b/core/src/main/scala/zio/config/ConfigDescriptor.scala
@@ -43,16 +43,16 @@ sealed trait ConfigDescriptor[K, V, A] { self =>
 
   def mapKey(f: K => K): ConfigDescriptor[K, V, A] = {
     def loop[B](config: ConfigDescriptor[K, V, B]): ConfigDescriptor[K, V, B] = config match {
-      case Source(path, source, propertyType) => Source(f(path), source, propertyType)
-      case Nested(path, conf)                 => Nested(f(path), loop(conf))
-      case Sequence(conf)                     => Sequence(loop(conf))
-      case Describe(config, message)          => Describe(loop(config), message)
-      case Default(value, value2)             => Default(loop(value), value2)
-      case Optional(config)                   => Optional(loop(config))
-      case XmapEither(config, f, g)           => XmapEither(loop(config), f, g)
-      case Zip(conf1, conf2)                  => Zip(loop(conf1), loop(conf2))
-      case OrElseEither(value1, value2)       => OrElseEither(loop(value1), loop(value2))
-      case OrElse(value1, value2)             => OrElse(loop(value1), loop(value2))
+      case Source(source, propertyType) => Source(source, propertyType)
+      case Nested(path, conf)           => Nested(f(path), loop(conf))
+      case Sequence(conf)               => Sequence(loop(conf))
+      case Describe(config, message)    => Describe(loop(config), message)
+      case Default(value, value2)       => Default(loop(value), value2)
+      case Optional(config)             => Optional(loop(config))
+      case XmapEither(config, f, g)     => XmapEither(loop(config), f, g)
+      case Zip(conf1, conf2)            => Zip(loop(conf1), loop(conf2))
+      case OrElseEither(value1, value2) => OrElseEither(loop(value1), loop(value2))
+      case OrElse(value1, value2)       => OrElse(loop(value1), loop(value2))
     }
     loop(self)
   }
@@ -73,16 +73,16 @@ sealed trait ConfigDescriptor[K, V, A] { self =>
 
   final def updateSource(f: ConfigSource[K, V] => ConfigSource[K, V]): ConfigDescriptor[K, V, A] = {
     def loop[B](config: ConfigDescriptor[K, V, B]): ConfigDescriptor[K, V, B] = config match {
-      case Source(path, source, propertyType) => Source(path, f(source), propertyType)
-      case Nested(path, conf)                 => Nested(path, loop(conf))
-      case Sequence(conf)                     => Sequence(loop(conf))
-      case Describe(config, message)          => Describe(loop(config), message)
-      case Default(value, value2)             => Default(loop(value), value2)
-      case Optional(config)                   => Optional(loop(config))
-      case XmapEither(config, f, g)           => XmapEither(loop(config), f, g)
-      case Zip(conf1, conf2)                  => Zip(loop(conf1), loop(conf2))
-      case OrElseEither(value1, value2)       => OrElseEither(loop(value1), loop(value2))
-      case OrElse(value1, value2)             => OrElse(loop(value1), loop(value2))
+      case Source(source, propertyType) => Source(f(source), propertyType)
+      case Nested(path, conf)           => Nested(path, loop(conf))
+      case Sequence(conf)               => Sequence(loop(conf))
+      case Describe(config, message)    => Describe(loop(config), message)
+      case Default(value, value2)       => Default(loop(value), value2)
+      case Optional(config)             => Optional(loop(config))
+      case XmapEither(config, f, g)     => XmapEither(loop(config), f, g)
+      case Zip(conf1, conf2)            => Zip(loop(conf1), loop(conf2))
+      case OrElseEither(value1, value2) => OrElseEither(loop(value1), loop(value2))
+      case OrElse(value1, value2)       => OrElse(loop(value1), loop(value2))
     }
     loop(self)
   }
@@ -122,7 +122,7 @@ object ConfigDescriptor {
 
   final case class Sequence[K, V, A](config: ConfigDescriptor[K, V, A]) extends ConfigDescriptor[K, V, List[A]]
 
-  final case class Source[K, V, A](path: K, source: ConfigSource[K, V], propertyType: PropertyType[V, A])
+  final case class Source[K, V, A](source: ConfigSource[K, V], propertyType: PropertyType[V, A])
       extends ConfigDescriptor[K, V, A]
 
   final case class Zip[K, V, A, B](left: ConfigDescriptor[K, V, A], right: ConfigDescriptor[K, V, B])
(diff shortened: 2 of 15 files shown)


afsalthaj:
Thanks @senia-psm for making the suggestion as a PR. Good one!
Will take a look.

afsalthaj:
The ergonomics is good now. list(string).at(path) instead of list(string(path)) which is actually a good win. Approving this. 

 I am still unsure how this is going to play out with typesafe config. The semantics is difficult to consume with respect to sources like map, properties, command line etc. We are adding the situation of users not giving a path for map source resulting in runtime exception. However, if we can make the typesafe config hocon work without a path, that will be infact a real value add, and that trade off is I think worth it.  

@jdegoes any suggestions here?

afsalthaj:
@senia-psm also could you add sbt runAllExamples as part of the build. We don’t need breaking examples. If there are any examples without assertions let’s change that as well :)

magnolia/src/main/scala/zio/config/magnolia/DeriveConfigDescriptor.scala
```diff
@@ -163,10 +163,10 @@ object DeriveConfigDescriptor {
         val finalDesc =
           if (caseClass.isObject) {
             val config = parentClass match {
-              case Some(parentClass) => string(parentClass.toLowerCase())
+              case Some(parentClass) => string.at(parentClass.toLowerCase())
 
               case None =>
-                string("").xmapEither[String](
+                (string at "").xmapEither[String](
```
afsalthaj, line 169: I think now we can remove `at ""` here

docs/refined/index.md
```diff
@@ -26,10 +26,10 @@ object RefinedReadConfig extends App {
 
   def prodConfig =
     (
-      nonEmpty(string("LDAP")) |@|
-        greaterEqual[W.`1024`.T](int("PORT")) |@|
-        nonEmpty(string("DB_URL")).optional |@|
-        size[Greater[W.`2`.T]](list(long("LONGVALS")))
+      nonEmpty(string.at("LDAP")) |@|
```
afsalthaj, line 29: Guess, I think we need some poofs now.
`nonEmtpy(string).at("LDAP") == nonEmpty(string.at("LDAP"))`

core/src/main/scala/zio/config/Config.scala
```diff
@@ -129,7 +129,7 @@ object Config {
    *
    * {{{
    *    final case class kafkaConfig(server: String, serde: String)
-   *    nested("KAFKA")(string("SERVERS") |@| string("SERDE"))(KafkaConfig.apply, KafkaConfig.unapply)
+   *    nested("KAFKA")(string.at("SERVERS") |@| string.at("SERDE"))(KafkaConfig.apply, KafkaConfig.unapply)
```
afsalthaj, line 132: Need some tests/proofs. Isn't it?
nested("KAFKA")(string.at("SERVERS"))  == string.at("SERVERS").at("KAFKA")

core/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -134,38 +136,38 @@ object ConfigDescriptor {
     g: B => Either[String, A]
   ) extends ConfigDescriptor[K, V, B]
 
-  def bigDecimal(path: String): ConfigDescriptor[String, String, BigDecimal] =
-    ConfigDescriptor.Source(path, ConfigSource.empty, PropertyType.BigDecimalType) ?? "value of type bigdecimal"
+  val bigDecimal: ConfigDescriptor[String, String, BigDecimal] =
+    ConfigDescriptor.Source(ConfigSource.empty[String, String], PropertyType.BigDecimalType) ?? "value of type bigdecimal"
```
afsalthaj, line 140: I hope `ConfigSource.empty` works?

core/src/main/scala/zio/config/ConfigDocs.scala
```diff
@@ -2,7 +2,7 @@ package zio.config
 
 sealed trait ConfigDocs[+K, +V]
 object ConfigDocs {
-  sealed trait Details[+V] {
+  sealed trait Details[+V] extends ConfigDocs[Nothing, V] {
```
afsalthaj, line 5: That's interesting. Why do we need to extend here?

senia-psm: It's a case on `Source`. Since `Source` has no path it produces `Details` directly.

core/src/main/scala/zio/config/ConfigDescriptor.scala
senia-psm: It works, but IDEA is complaining. I'll revert it.

core/src/main/scala/zio/config/Config.scala
senia-psm: > @senia-psm also could you add sbt runAllExamples as part of the build. We don’t need breaking examples. If there are any examples without assertions let’s change that as well :)

I'm getting `Nonzero exit code returned from runner: 1` without any details on `sbt runAllExamples`

senia-psm:
> @senia-psm also could you add sbt runAllExamples as part of the build. We don’t need breaking examples. If there are any examples without assertions let’s change that as well :)
I'm getting `Nonzero exit code returned from runner: 1` without any details on `sbt runAllExamples`.

core/src/main/scala/zio/config/Config.scala
senia-psm: All tests are already reimplemented with `.at`. I have no idea what should I test here.

magnolia/src/main/scala/zio/config/magnolia/DeriveConfigDescriptor.scala
senia-psm: done

afsalthaj:
@senia-psm  Try the same task in local.

sbt 
 > runAllExamples

senia-psm:
> @senia-psm Try the same task in local.
> 
> sbt
> 
> > runAllExamples

you mean in `sbt` console? Same result.

afsalthaj:
@senia-psm  Hey! Can you please take a look at sbt runAllExamples?
It runs the examples and if assertion fail, you should get an assertion-failed exception.

Or run the examples individually. See which one fails.
sbt >
    > project zio-config-examples
    > run

docs/refined/index.md
senia-psm: done

senia-psm:
> Or run the examples individually. See which one fails.

`SimpleExampleMain` fails due to missing file.

All other examples are fine.

I'm removing `WIP`.

afsalthaj:
On SimpleExampleMain,, let's put a resource file. and make that work as well.

core/src/main/scala/zio/config/Config.scala
afsalthaj: The idea is, now users have two ways in which they form `nested` paths, unlike before.

So let's add a simple test, which is
```scala
nested("KAFKA")(string.at("SERVERS")) == string.at("SERVERS").at("KAFKA")


```
also worth testing
```scala
list(string.at("path")) = list(string).at("path")
```

afsalthaj: The reason is, currently we have two ways in which some one can list. Conceptually, this was not the case before. The only way to create a list, is , for instance:

```sala
list(string("path"))
```

This is why I think, worth adding a test

senia-psm:
@afsalthaj we can address `SimpleExampleMain` in a separate PR. I have no idea what is expected content of this file. See #260.

core/src/main/scala/zio/config/ConfigDocs.scala
afsalthaj: I am hoping this can be made part of the ADT, just like the other ones.

 `case class Details(sources: Option[Sources], descriptions: List[String]) extends ConfigDoc[A] `

This can also exist with sources not being an `Option`, because it is anyways a list, and we can get away with it by 
 saying `Sources(Nil)` 

// And may be for some readability:
case class Sources(list: List[String])

object Sources {
  def empty = Sources(Nil)
}

afsalthaj: There is a usage of `DescriptionsWithValue`, I think, may be we can have one more field with `Option[V]` and avoid `Descriptions` and `DescriptionWithValue`, and just `Details` for everything

core/src/main/scala/zio/config/ConfigDocsFunctions.scala
```diff
@@ -63,31 +60,39 @@ private[config] trait ConfigDocsFunctions {
...
+              Both(loop(tree, flattened, left, initialValue), loop(tree, flattened, right, initialValue))
 
             case OneOf(left, right) =>
-              OneOf(loop(left, initialValue), loop(right, initialValue))
+              OneOf(loop(tree, flattened, left, initialValue), loop(tree, flattened, right, initialValue))
+
+            case Sequence(element) =>
+              tree.getPath(initialValue.toList) match {
+                case PropertyTree.Sequence(value) if value.nonEmpty =>
+                  SequenceWithValue(value.map(t => loop(t, t.flatten, element, Vector.empty)))
```
afsalthaj, line 89: With the new Details, I think, `SequenceWithValue` can be removed, and have Sequence as `Sequence(list: List[ConfigDocs])` and provide a Sequence(List(...)) if the loop produces only one entry.

afsalthaj:
@senia-psm  I think, most of the comments are minor changes. its LGTM overall.

afsalthaj:
@senia-psm  On SimpleExampleMain, leave it to me. I will fix it in another PR

core/src/main/scala/zio/config/ConfigDocs.scala
```diff
@@ -15,8 +15,9 @@ object ConfigDocs {
   final case class Sources(list: List[String])
 
   final case object Empty                                                       extends ConfigDocs[Nothing, Nothing]
-  final case class Path[K, V](path: K, details: Details[V])                     extends ConfigDocs[K, V]
   final case class NestedPath[K, V](path: K, docs: ConfigDocs[K, V])            extends ConfigDocs[K, V]
   final case class Both[K, V](left: ConfigDocs[K, V], right: ConfigDocs[K, V])  extends ConfigDocs[K, V]
   final case class OneOf[K, V](left: ConfigDocs[K, V], right: ConfigDocs[K, V]) extends ConfigDocs[K, V]
+  final case class Sequence[K, V](element: ConfigDocs[K, V])                    extends ConfigDocs[K, V]
+  final case class SequenceWithValue[K, V](element: List[ConfigDocs[K, V]])     extends ConfigDocs[K, V]
```
afsalthaj, line 22: case class Details(v: Option[V], sources: Sources, descriptions; List[String]) extends ConfigDocs[K, V]
.....
case class Sequence(element: List[ConfigDocs[K, V]]) extends ConfigDocs[K, V]

afsalthaj: Pretty printing this is going to be a bit tricky/non-trivial. 
Not for this PR, but worth mentioning about it, before we hit 1.0

senia-psm: done

core/src/main/scala/zio/config/ConfigDocsFunctions.scala
senia-psm: done

core/src/main/scala/zio/config/ConfigDocs.scala
senia-psm: done

core/src/main/scala/zio/config/Config.scala
senia-psm: Actually no, right now `list(string.at("path")) = list(string).at("path")`, but I'm going to break it in #261 since there is a bug related to it. I mentioned this bug earlier in discord.
I'm going to enforce this kind of behavior to fix #261:
`list(string).at("path")` <-> `{path: ["a"]}`
`list(string.at("path"))` <-> `[{path: "a"}]`

afsalthaj:
A point to note here:


With the new change (i.e, descriptor without a path), the implications and case-by-case handling increases, leading to more scenarios to be thought about.
```
list(string).at("path")               <-> {path: ["s"]}
list(string.at("path").optional) <-> [{}]
list(string.at("path"))               <-> [{path: "s"}]

```
I am not saying its wrong, It's ok enough. But we should verify, what is the real value add for the ussers.

I think, one of the main value-add that I found with being able to write `list(string).at("path")` is the readability over `list(string("path"))`

To achieve this readability, I think its much more easier. Please find the following code:

```scala

val string: String => ConfigDescriptor[ String] =
    ConfigDescriptor.Source(_, ConfigSource.empty, PropertyType.StringType) ?? " string"

def list(f: String => ConfigDescriptor[String, String, String]): String => ConfigDescriptor[String] = 
  path =>  Sequence[K, V, A](f(path)

implicit class Ops[A](f: String => ConfigDescriptor[A]) {
   def at(path: String) = f(path)
}

```

```scala

list(string).at("path") // and in fact that's the only way!

// and others will work as it is
string("path")

```

senia-psm:
@afsalthaj Unfortunately the described behavior you have described is inconsistent and too complicated.

## Inconsistency1
What is the correct description for these 2 configs?
1.
```
a: {
  b: ["str"]
}
```
2.
```
a: [{ b: "str" }]
```
In current behavior it's unclear. There is no reasonable explanation why `nested("a")(list(string("b")))` means `{a: { b: ["str"] }}`, but not `{a: [{b: "str"}]}`.
## Inconsistency1:
If `nested("a")(list(string("b")))` means `{a: { b: ["str"] }}`, then what is the meaning of `nested("a")(list(string("b").orElseEither(int("c"))))`.
In your case... I have no idea.
In my case it should be:
```
a: [{ b: "str" }, { c: 1 }]
```.
##Too complicated
Please see this message: https://github.com/zio/zio-config/issues/254#issuecomment-609754662
Current implementation breaks documentation together with parsing of empty lists in some cases.
You can find test case in #261.

**Comment by senia-psm** [comment]
@afsalthaj regarding this approach: `def list(f: String => ConfigDescriptor[String, String, String])`

It breaks lists of objects:
```scala
val cInnerConfig = (string.at("a") |@| string.at("b"))(Inner, Inner.unapply)

val cListConfig = list(cInnerConfig).at("list")(Outer, Outer.unapply)
```
And, in fact, gives you nothing:
```scala
list(x => string(x).orElseEither(int("b")))
```
```scala
def cInnerConfig(path: String) = (string.at("a") |@| string.at("b"))(Inner, Inner.unapply)

val cListConfig = list(cInnerConfig).at("list")(Outer, Outer.unapply)
```
So the ergonomics is war from perfect and it's error prone.

afsalthaj:
> @afsalthaj Unfortunately the described behavior you have described is inconsistent and too complicated.

```
list(string("a")) ==> for a key `a` there is a list of string
```

for the same reasons
```
nested("a)(list(string("b")) ==> `inside a, there exists a key `b`, we have a list of string

```

for the same reasons

```
nested("a")(list(string("b").orElseEither(int("c"))))  ==> 
  inside a , we can have either of the following:
    1. a key `b` whose value is a list of string
    2. a key `c` whose value is integer
```

I understand your preference, but let's don't tag it as right and wrong

senia-psm:
> ```
> nested("a")(list(string("b").orElseEither(int("c"))))  ==> 
>   inside a , we can have either of the following:
>     1. a key `b` whose value is a list of string
>     2. a key `c` whose value is integer
> ```

Are you sure about that? `int` is inside of `list` in Description, but outside of list in config? Sorry, but I have to tag it as wrong.

afsalthaj:
> It breaks lists of objects:

> val cInnerConfig = (string.at("a") |@| string.at("b"))(Inner, Inner.unapply)

> val cListConfig = list(cInnerConfig).at("list")(Outer, Outer.unapply)

Simply use the current solution:

```scala
nested("list") { cInnerConfig } 
```

Inside `list`,there is a list of `{ a : [string] , b : [String }`

afsalthaj:
> Are you sure about that? int is inside of list in Description, but outside of list in config? Sorry, but > I have to tag it as wrong.

I have edited the answer. Also I think that makes sense

A similar scenario is depicted as an example here:
https://github.com/zio/zio-config/blob/master/examples/src/main/scala/zio/config/examples/typesafe/NullAndOptionalConfig.scala

senia-psm:
> Simply use the current solution:

I know the workaroung, but it's both counterintuitive for users and allows lots of invalid, but compilable cases like I described above.

afsalthaj:
Also, let's keep this in a branch. I don't think its a candidate for 1.0

afsalthaj:
> I know the workaroung, but it's both counterintuitive for users and allows lots of invalid, but compilable cases like I described above.

I think its a preference at this stage. Personally I really liked the outcome  `list(string).at("path")`.
However, there are many who like list(string("path")). It depends on users. We can get feedback. We have enough user-base for a feedback now. So let's wait.

Also, the solution in this PR leads to
`list(string) from ConfigSource.fromMap(map)` or 
`list(string) from sysEnvSource`

These don't make sense at compile time even

senia-psm:
> ```
> nested("a")(list(string("b").orElseEither(int("c")))  ==> 
>   inside a , we can have a list of either of the following:
>     1. a key `b` whose value is a string
>     2. a key `c` whose value is integer
> ```
To get the consistent `<*>` behavior are you going to require both `b` as list of strings and `c` as list of integers and then zip them together?

afsalthaj:
Also, I think we already give users a strong/complicated example,

https://github.com/zio/zio-config/blob/master/examples/src/main/scala/zio/config/examples/typesafe/NullAndOptionalConfig.scala

That should give some idea. A big change at this stage for 1.0 can be a bit risky as well

senia-psm:
> These don't make sense at compile time even

Actually it makes sense:
```scala
val cList = list(string)

val actual = read(
  cList from ConfigSource
    .fromPropertyTree(PropertyTree.Sequence(List(PropertyTree.Leaf("a"), PropertyTree.Leaf("b"))), "seq")
)

assert(actual)(isRight(equalTo(List("a", "b"))))
```
And with `fromMap` - it's just a limitation of `Map`, not `zio-config` itself.

senia-psm:
> That should give some idea. A big change at this stage for 1.0 can be a bit risky as well

I can't get your point.
`nested("employees")(list(employee))` (from `NullAndOptionalConfig`) should work as expected.
On the other hand `list(employee.at(employees"))` would produce a different result.

And, by the way, `nested("employees")(list(employee))` fails on empty lists with cryptic error message.

senia-psm:
> That should give some idea. A big change at this stage for 1.0 can be a bit risky as well

Actually my approach keeps `nested("employees")(list(employee))` (from `NullAndOptionalConfig`) intact and your approach with `def list(f: String => ConfigDescriptor[String, String, String])` breaks it. An it requires lots and lots of changes (including `ProductBuilder`) to keep `NullAndOptionalConfig` as is. And such changes can break type inference in IDEA.

afsalthaj:
@leigh-perry  Do y

afsalthaj:
Closing this as PR, and let's keep the branch.

senia-psm:
@afsalthaj what is the reason for closing this PR?

jdegoes:
@senia-psm @afsalthaj 

Here's what I suggest:

First off, I like the idea to have non-record types. Right now, `string("foo")`, etc., are all record types. This prevents forming a list type with a specified (anonymous) element type, as well as a map type, with a specified (anonymous) value type for the values in the map—both of which are useful, whether or not they are supported by HOCON.

So this is a positive change, we just need to figure out how to do it best.

The `at` machinery complicates things considerably for the common case, so I would not recommend it. Rather, I would recommend the following:

For each property type, have two constructors:

```scala
val string: ConfigDescriptor[String, String] = ...
def string(name: String): ConfigDescriptor[String, String] = nested(name)(string)
```

This way, existing code will not break unnecessarily, the concise syntax of the common case is retained, and both use cases are satisfied.

I will make other comments on the code itself.

core/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -32,6 +32,8 @@ sealed trait ConfigDescriptor[K, V, A] { self =>
   final def <+>[B](that: => ConfigDescriptor[K, V, B]): ConfigDescriptor[K, V, Either[A, B]] =
     self orElseEither that
 
+  final def at(k: K): ConfigDescriptor[K, V, A] = ConfigDescriptor.nested(k)(this)
```
jdegoes, line 35: I would delete this operator as it will no longer be necessary after other changes I proposed.

core/src/main/scala/zio/config/Config.scala
```diff
@@ -161,7 +161,7 @@ object Config {
...
```
jdegoes, line 164: All these examples should survive with no changes to syntax.

core/src/main/scala/zio/config/ConfigDescriptor.scala
senia-psm: Done. Though I'd prefer to keep this method: `nested(path)(inner)` looks too verbose for me.

core/src/main/scala/zio/config/Config.scala
senia-psm: All docs reverted.

senia-psm:
@afsalthaj @jdegoes when can I expect this PS to get merged? I have other fixes that depend on this PR and it's quite troublesome to constantly update those branches after review.

jdegoes:
@senia-psm Assuming it passes, I think it's now good to merge. Big improvement to readability and backward compatibility!

It'd be nice to add a `list("foo")(cd)` variation for consistency but that could be done in a separate PR.

jdegoes:
@senia-psm Nice work! Big improvement. Thanks for your work on this.

Issue #254, opened 2020-04-06
Invalid derived ConfigDescriptor for nested sealed traits


senia-psm:
Version: 1.0.0-RC15

**Classes:**
```scala
sealed trait X
case object A extends X
case object B extends X

case class Cfg(fieldName: X)
```

**Expected behavior:**
This configuration should produce an instance of `Cfg` when parsed with `DeriveConfigDescriptor[Cfg]`
```
fieldName {
  x = a
}
```

**Actual behavior:**
Error produced:
```
Left(OrErrors(List(MissingValue(Vector(Right(fieldName), Right(X), Right(fieldName))), MissingValue(Vector(Right(fieldName), Right(X), Right(fieldName))))))
```

For some reason expected config string is:
```
fieldName {
  X { fieldName = a }
}
```
Note upper case `X` and `fieldName` inside `X`.

afsalthaj:
Looking at it now.

afsalthaj:
@senia-psm 
For the time being, if you are blocked, please use

```scala

  val desc = nested("fieldName") {
    DeriveConfigDescriptor[X]
  }

```

I will be fixing the issue soon.

senia-psm:
@afsalthaj thank you!
Please note that there are several issues related to `Coproduct`.
For instance:
Correct:
```scala
case class CfgL1(fieldName: List[X])
write(DeriveConfigDescriptor[CfgL1], CfgL1(A :: Nil)).toString ==
  "Right(Sequence(List(Record(Map(fieldName -> Record(Map(X -> Record(Map(fieldName -> Leaf(a))))))))))"
```
Incorrect:
```scala
case class CfgL2(s: String, fieldName: List[X])
write(DeriveConfigDescriptor[CfgL2], CfgL2("h", A :: Nil)).toString ==
  "Right(Sequence(List(Record(Map(fieldName -> Record(Map(X -> Record(Map(fieldName -> Empty)))), s -> Empty)))))"
```

Another issue:
```scala
val aHoconSource = TypeSafeConfigSource.fromHoconString("fieldName = []").getOrElse(sys.error("Err"))
read(DeriveConfigDescriptor[CfgL2] from aHoconSource)
// Left(OrErrors(List(MissingValue(Vector(Right(fieldName), Right(X), Right(fieldName))), MissingValue(Vector(Right(fieldName), Right(X), Right(fieldName))))))
```

Should I create all these issues or wait for this issue fix and retest?

PR #261, opened 2020-04-07
Fix empty list processing

senia-psm:
`ReadFunctions` is significantly simplified.

`ReadFunctions` is simpler and more strict now. All existing use-cases (except for one) supported with combined `ConfigDescriptor`s by default.

## Main changes
- `ConfigDescriptor.Sequence` holds `ConfigSource`.
- `ConfigDescriptor.Source` can read only `Leaf` node.
- `ConfigDescriptor.Sequence` can read only `Sequence` node.
- No more errors with empty list parsing.
- We can use `List` as a first class value including `Option[List[T]]`, `List[List[T]]`, default values, etc.
- Results of `ConfigSource.fromMap` and `ConfigSource.fromCommandLineArgs` are simplified.
- We can distinguish scalar and list.

## Support for existing use cases
- `list(nested(path)(cfg))` is transformed to `list(path)(cfg)` under the hood.
- `list(path)(cfg)` accepts scalar value as a list, i.e. `list("a")(string)` can parse both `a = ["v"]` and `a = "v"`. This is used in tests for `fromCommandLineArgs`.

## Broken use case
`string(path)` won't accept list by default. There is method `first` that consumes both scalar value and lists and returns first value: `first(path)(string)`

## `Either[String, List[String]]`
We can distinguish scalars and lists.
Now both `Either[String, List[String]]` and `Either[List[String], String]` make sense. On `master` any list descriptor can consume scalar and any scalar descriptor can consume list.
In this PR scalar descriptors can't consume lists by default and there is a strict version of list: `listStrict`.

## Corner cases
Nested lists are supported. We can even parse lists with empty list inside ( `path = [[]]` ) as `List(Nil)` and it's quite easy to create a parser for nested lists with single field objects: `path = [[{a = "b"}]]`.
For list of corner cases with descriptors see `ListsCornerCasesTest.scala`.

## Magnolia
`zio-config-magnolia` support for `Option[NonEmptyList[T]]` is fixed! It won't assumes that `descr.optional` can recover from `ConversionError` any more.
**Reason:** `zio-config-refined` assumes that `descr.optional` can't recover from `ConversionError`.
I suspect both work on `master` due to inconsistency in `descr.optional` behavior with lists.

This PR fixes `List` support, so there is no need for `Option[NonEmptyList[T]]` support any way, but we still can have it.

### DIFF
--- a/core/src/main/scala/zio/config/ConfigDescriptor.scala
+++ b/core/src/main/scala/zio/config/ConfigDescriptor.scala
@@ -178,9 +178,14 @@ object ConfigDescriptor {
   val float: ConfigDescriptor[String, String, Float] =
     ConfigDescriptor.Source(ConfigSource.empty, PropertyType.FloatType) ?? "value of type float"
 
-  def first[K, V, A](path: K)(desc: ConfigDescriptor[K, V, A]) =
-    list(path)(desc)
-      .xmapEither[A](_.headOption.fold[Either[String, A]](Left("Element is missing"))(Right(_)), v => Right(v :: Nil))
+  def head[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, A] =
+    desc.orElse(
+      listStrict(desc)
+        .xmapEither[A](_.headOption.fold[Either[String, A]](Left("Element is missing"))(Right(_)), v => Right(v :: Nil))
+    )
+
+  def head[K, V, A](path: K)(desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, A] =
+    nested(path)(head(desc))
 
   def float(path: String): ConfigDescriptor[String, String, Float] = nested(path)(float)
 
@@ -190,54 +195,28 @@ object ConfigDescriptor {
   def int(path: String): ConfigDescriptor[String, String, Int] = nested(path)(int)
 
   /**
-   * Leaks out path in simple cases, i.e.
-   * `list(string(path))` == `list(path)(string)` == `nested(path)(listOrSingle(string))`
-   *
-   * `nested("a")(list(string("b")))` describes configuration `{a: { b: ["s1", "s2"] }}`
-   *
    * Allows scalar value instead of list
    * */
-  def list[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] = {
-    def extractPath(cfg: ConfigDescriptor[K, V, A]): Option[(K, ConfigDescriptor[K, V, A])] = cfg match {
-      case Describe(config, message) =>
-        extractPath(config).map { case (path, inner) => (path, Describe(inner, message)) }
-      case Nested(path, config) => Some((path, config))
-      case _                    => None
-    }
-
-    extractPath(desc) match {
-      case Some((path, inner)) => list(path)(inner)
-      case None                => listOrSingle(desc)
-    }
-  }
-
-  /**
-   * Keeps inner description intact.
-   *
-   * Allows scalar value instead of list
-   * */
-  def list[K, V, A](path: K)(desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
-    nested(path)(listOrSingle(desc))
-
-  /**
-   * Keeps inner description intact.
-   *
-   * `nested("a")(listOrSingle(string("b")))` describes configuration `{a: [{b: "s1"}, {b: "s2"}]}`
-   *
-   * Allows scalar value instead of list
-   * */
-  def listOrSingle[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
+  def list[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
     listStrict(desc).orElse(desc(_ :: Nil, _.headOption))
 
   /**
-   * Keeps inner description intact.
-   *
-   * `nested("a")(strictList(string("b")))` describes configuration `{a: [{b: "s1"}, {b: "s2"}]}`
-   *
+   * Allows scalar value instead of list
+   * */
+  def list[K, V, A](path: K)(desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
+    nested(path)(list(desc))
+
+  /**
    * Rejects scalar value in place of list
    * */
   def listStrict[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
     ConfigDescriptor.Sequence(ConfigSource.empty, desc)
+
+  /**
+   * Rejects scalar value in place of list
+   * */
+  def listStrict[K, V, A](path: K)(desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
+    nested(path)(listStrict(desc))
 
   val long: ConfigDescriptor[String, String, Long] =
     ConfigDescriptor.Source(ConfigSource.empty, PropertyType.LongType) ?? "value of type long"
--- a/core/src/main/scala/zio/config/ConfigDescriptor.scala
+++ b/core/src/main/scala/zio/config/ConfigDescriptor.scala
@@ -47,7 +47,7 @@ sealed trait ConfigDescriptor[K, V, A] { self =>
     def loop[B](config: ConfigDescriptor[K, V, B]): ConfigDescriptor[K, V, B] = config match {
       case Source(source, propertyType) => Source(source, propertyType)
       case Nested(path, conf)           => Nested(f(path), loop(conf))
-      case Sequence(conf)               => Sequence(loop(conf))
+      case Sequence(source, conf)       => Sequence(source, loop(conf))
       case Describe(config, message)    => Describe(loop(config), message)
       case Default(value, value2)       => Default(loop(value), value2)
       case Optional(config)             => Optional(loop(config))
@@ -77,7 +77,7 @@ sealed trait ConfigDescriptor[K, V, A] { self =>
     def loop[B](config: ConfigDescriptor[K, V, B]): ConfigDescriptor[K, V, B] = config match {
       case Source(source, propertyType) => Source(f(source), propertyType)
       case Nested(path, conf)           => Nested(path, loop(conf))
-      case Sequence(conf)               => Sequence(loop(conf))
+      case Sequence(source, conf)       => Sequence(f(source), loop(conf))
       case Describe(config, message)    => Describe(loop(config), message)
       case Default(value, value2)       => Default(loop(value), value2)
       case Optional(config)             => Optional(loop(config))
@@ -122,7 +122,8 @@ object ConfigDescriptor {
   final case class OrElseEither[K, V, A, B](left: ConfigDescriptor[K, V, A], right: ConfigDescriptor[K, V, B])
       extends ConfigDescriptor[K, V, Either[A, B]]
 
-  final case class Sequence[K, V, A](config: ConfigDescriptor[K, V, A]) extends ConfigDescriptor[K, V, List[A]]
+  final case class Sequence[K, V, A](source: ConfigSource[K, V], config: ConfigDescriptor[K, V, A])
+      extends ConfigDescriptor[K, V, List[A]]
 
   final case class Source[K, V, A](source: ConfigSource[K, V], propertyType: PropertyType[V, A])
       extends ConfigDescriptor[K, V, A]
@@ -177,6 +178,10 @@ object ConfigDescriptor {
   val float: ConfigDescriptor[String, String, Float] =
     ConfigDescriptor.Source(ConfigSource.empty, PropertyType.FloatType) ?? "value of type float"
 
+  def first[K, V, A](path: K)(desc: ConfigDescriptor[K, V, A]) =
+    list(path)(desc)
+      .xmapEither[A](_.headOption.fold[Either[String, A]](Left("Element is missing"))(Right(_)), v => Right(v :: Nil))
+
   def float(path: String): ConfigDescriptor[String, String, Float] = nested(path)(float)
 
   val int: ConfigDescriptor[String, String, Int] =
@@ -184,8 +189,55 @@ object ConfigDescriptor {
 
   def int(path: String): ConfigDescriptor[String, String, Int] = nested(path)(int)
 
-  def list[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
-    ConfigDescriptor.Sequence(desc)
+  /**
+   * Leaks out path in simple cases, i.e.
+   * `list(string(path))` == `list(path)(string)` == `nested(path)(listOrSingle(string))`
+   *
+   * `nested("a")(list(string("b")))` describes configuration `{a: { b: ["s1", "s2"] }}`
+   *
+   * Allows scalar value instead of list
+   * */
+  def list[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] = {
+    def extractPath(cfg: ConfigDescriptor[K, V, A]): Option[(K, ConfigDescriptor[K, V, A])] = cfg match {
+      case Describe(config, message) =>
+        extractPath(config).map { case (path, inner) => (path, Describe(inner, message)) }
+      case Nested(path, config) => Some((path, config))
+      case _                    => None
+    }
+
+    extractPath(desc) match {
+      case Some((path, inner)) => list(path)(inner)
+      case None                => listOrSingle(desc)
+    }
+  }
+
+  /**
+   * Keeps inner description intact.
+   *
+   * Allows scalar value instead of list
+   * */
+  def list[K, V, A](path: K)(desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
+    nested(path)(listOrSingle(desc))
+
+  /**
+   * Keeps inner description intact.
+   *
+   * `nested("a")(listOrSingle(string("b")))` describes configuration `{a: [{b: "s1"}, {b: "s2"}]}`
+   *
+   * Allows scalar value instead of list
+   * */
+  def listOrSingle[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
+    listStrict(desc).orElse(desc(_ :: Nil, _.headOption))
+
+  /**
+   * Keeps inner description intact.
+   *
+   * `nested("a")(strictList(string("b")))` describes configuration `{a: [{b: "s1"}, {b: "s2"}]}`
+   *
+   * Rejects scalar value in place of list
+   * */
+  def listStrict[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
+    ConfigDescriptor.Sequence(ConfigSource.empty, desc)
 
   val long: ConfigDescriptor[String, String, Long] =
     ConfigDescriptor.Source(ConfigSource.empty, PropertyType.LongType) ?? "value of type long"
--- a/core/src/main/scala/zio/config/ConfigSource.scala
+++ b/core/src/main/scala/zio/config/ConfigSource.scala
@@ -78,7 +78,7 @@ object ConfigSource {
    *
    *    final case class Config(databaseCredentials: Credentials, vaultCredentials: Credentials, regions: List[String], users: List[String])
    *
-   *    (nested("db") { credentials } |@| nested("vault") { credentials } |@| list(string("regions")) |@| list(string("user")))(Config.apply, Config.unapply)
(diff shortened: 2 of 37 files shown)


afsalthaj:
What is the older bug?

afsalthaj:
There was only behaviour before in zio-config.
This is as follows:

```
given:
val config = list(string("path"))

```
```
config <-> {path: ["s"]}

```

```

nested(parent)(config) <-> {path: ["s"]}

```

The only time this became verbose is for coproducts. Which is a completely separate discussion.



-------------------------------------------

With the new change, which is adding an `at` function, the implications and case-by-case handling increases, leading to more thoughts and complications in the library. You have already given some examples in the PR description. This leads to more scenarios

```
`list(string).at("path") <-> {path: ["s"]} ==> list(string.at("path").optional) <-> [{}]

```
. I am not saying its wrong, It's ok enough. 
I think main value-add that I found with being able to write `list(string).at("path")` is the readability over `list(string("path"))`

To achieve this readability, I think its quite easy. It's a matter of doing something like below:

```scala

val string: String => ConfigDescriptor[ String] =
    ConfigDescriptor.Source(_, ConfigSource.empty, PropertyType.StringType) ?? " string"

def list(f: String => ConfigDescriptor[String, String, String]): String => ConfigDescriptor[String] = 
  path =>  Sequence[K, V, A](f(path)

implicit class Ops[A](f: String => ConfigDescriptor[A]) {
   def at(path: String) = f(path)
}

```

```scala

list(string).at("path") // and in fact that's the only way!

// and others will work as it is
string("path")

```

senia-psm:
> What is the older bug?

Please see this message for bug description (broken parsing of empty list): https://github.com/zio/zio-config/issues/254#issuecomment-609754662

senia-psm:
@afsalthaj Please see this message regarding you comment on approach: https://github.com/zio/zio-config/pull/259#issuecomment-610232476

afsalthaj:
@senia-psm thanks for your work. If this is a WIP, let’s keep the branch and close the PR until you are ready to raise again?

senia-psm:
@afsalthaj @jdegoes can we merge this PR with broken NonEmptyList support in magnolia (see details in PR description) and address it in a separate PR?
I don't think I can fix magnolia in a reasonable time frame.

This PR fixes `List` support, so there is no need in `Option[NonEmptyList[T]]` tricks any way.

senia-psm:
> @senia-psm thanks for your work. If this is a WIP, let’s keep the branch and close the PR until you are ready to raise again?

Not a `WIP` any more.

senia-psm:
@afsalthaj @jdegoes I've managed to fix `zio-config-magnolia` support for `Option[NonEmptyList[T]]`! It won't assumes that `descr.optional` can recover from `ConversionError` any more.
All tests are green, all examples status codes are `0`.
There should be no issues with this PR any more.
Please review it.

Ant, btw, is here any chance to get a new version of `zio-config` with this fix soon? Broken empty list support block my project.

afsalthaj:
@senia-psm Thanks for the contribution again.

The PR does seem to clean up a few things, and looks good skimming through it.

While I review, do you mind updating the version of magnolia and see if it works ? (scala-steward PRs is failing)

Also, don't worry about release timings. It will be a quick release once merged.

senia-psm:
@afsalthaj Ok, I'll check magnolia update.

typesafe/src/test/scala/zio/config/typesafe/TypesafeConfigSpec.scala
```diff
@@ -1,16 +1,30 @@
...
+                |a {
+                |  b = "s"
+                |  c = []
+                |}
+                |""".stripMargin
+            )
+
+          val expected = Record(Map("a" -> Record(Map("b" -> Leaf("s"), "c" -> Sequence(Nil)))))
+
+          assert(res.map(_.getConfigValue(List.empty)))(isRight(equalTo(expected)))
```
afsalthaj, line 26: 👍

afsalthaj: Could we change the test just below this one and change some of the values to empty list (at different places) and assert this is true in all cases?

This is fantastic !

core/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -170,15 +171,66 @@ object ConfigDescriptor {
...
   val int: ConfigDescriptor[String, String, Int] =
     ConfigDescriptor.Source(ConfigSource.empty, PropertyType.IntType) ?? "value of type int"
 
   def int(path: String): ConfigDescriptor[String, String, Int] = nested(path)(int)
 
-  def list[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
-    ConfigDescriptor.Sequence(desc)
+  /**
+   * Leaks out path in simple cases, i.e.
+   * `list(string(path))` == `list(path)(string)` == `nested(path)(listOrSingle(string))`
```
jdegoes, line 187: I'm not sure these equalities should hold.

This config descriptor:

```
[foo = 2, foo = 3]
```

is different than this one:

```
foo = [2, 3]
```

and we should preserve this distinction.

senia-psm: `list` keeps backward compatibility with all the docks. See `listStrict` below for.

core/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -170,15 +171,66 @@ object ConfigDescriptor {
...
+    nested(path)(listOrSingle(desc))
+
+  /**
+   * Keeps inner description intact.
+   *
+   * `nested("a")(listOrSingle(string("b")))` describes configuration `{a: [{b: "s1"}, {b: "s2"}]}`
+   *
+   * Allows scalar value instead of list
+   * */
+  def listOrSingle[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
```
jdegoes, line 222: By consistency with previous conventions, this one should be called `list` only, because the above `list` with `path` just delegates to this one, calling `nested` to wrap it into the specified field of a map.

```suggestion
  def list[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
```

core/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -170,15 +171,66 @@ object ConfigDescriptor {
...
+    listStrict(desc).orElse(desc(_ :: Nil, _.headOption))
+
+  /**
+   * Keeps inner description intact.
+   *
+   * `nested("a")(strictList(string("b")))` describes configuration `{a: [{b: "s1"}, {b: "s2"}]}`
+   *
+   * Rejects scalar value in place of list
+   * */
+  def listStrict[K, V, A](desc: ConfigDescriptor[K, V, A]): ConfigDescriptor[K, V, List[A]] =
```
jdegoes, line 232: I like the "forgiving" variant by default that accepts a single value.

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -16,28 +16,37 @@ import zio.Task
...
-      k => getConfigValue(k).getOrElse(that.getConfigValue(k)),
-      if (self.sourceDescription == that.sourceDescription) self.sourceDescription
-      else self.sourceDescription ++ that.sourceDescription
-    )
+sealed trait ConfigSource[K, V] { self =>
+  def sourceDescription: List[String]
+
+  def getConfigValue(path: List[K]): PropertyTree[K, V]
+
+  final def orElse(that: ConfigSource[K, V]): ConfigSource[K, V] = MergedConfigSource(self, that)
```
jdegoes, line 24: To preserve laziness:

```suggestion
  final def orElse(that: => ConfigSource[K, V]): ConfigSource[K, V] = MergedConfigSource(() => self, () => that)
```

core/src/main/scala/zio/config/ConfigDescriptor.scala
senia-psm: @jdegoes unfortunately to preserve current behavior of `list(string(path))` we need `list` to leak out nested path.

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -16,28 +16,37 @@ import zio.Task
...
-  final def <>(that: => ConfigSource[K, V]): ConfigSource[K, V] =
-    self orElse that
+  final def <>(that: => ConfigSource[K, V]): ConfigSource[K, V] = self orElse that
+}
+
+final case class TreeConfigSource[K, V](tree: PropertyTree[K, V], sourceDescription: List[String])
+    extends ConfigSource[K, V] {
+  def getConfigValue(path: List[K]): PropertyTree[K, V] = tree.getPath(path)
+}
+final case class MergedConfigSource[K, V](left: ConfigSource[K, V], right: ConfigSource[K, V])
```
jdegoes, line 33: It looks like this exists mainly to avoid accumulating duplicate source descriptions.

I'd suggest that `sourceDescription` should really be `Set[String]` to ensure there can be no duplicates, during composition or otherwise.

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -16,28 +16,37 @@ import zio.Task
...
+    extends ConfigSource[K, V] {
+  def getConfigValue(path: List[K]): PropertyTree[K, V] =
+    left.getConfigValue(path).getOrElse(right.getConfigValue(path))
+
+  val sourceDescription: List[String] =
+    if (left.sourceDescription == right.sourceDescription) left.sourceDescription
```
jdegoes, line 39: This logic isn't quite correct to fully prevent duplicates because it's order-sensitive. See above suggestion

core/src/main/scala/zio/config/ConfigSource.scala
```diff
@@ -296,28 +305,53 @@ object ConfigSource {
...
-      unflatten(
-        map.map(
-          tuple => {
-            val vectorOfKeys = keyDelimiter match {
-              case Some(keyDelimiter) => tuple._1.split(keyDelimiter).toVector.filterNot(_.trim == "")
-              case None               => Vector(tuple._1)
-            }
-            vectorOfKeys -> f(tuple._2)
-          }
+      unwrapSingletonLists(
```
jdegoes, line 308: What's the point of `unwrapSingletonLists` and `dropEmpty`?

core/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -75,16 +77,18 @@ sealed trait PropertyTree[+K, +V] { self =>
...
     k match {
       case Nil => self
       case head :: next =>
         self match {
-          case Empty         => Empty
-          case Leaf(_)       => Empty
-          case Record(value) => value.get(head.asInstanceOf[K]).map(_.getPath(next)).getOrElse(Empty)
-          case Sequence(r)   => Sequence(r.map(_.getPath(k)))
-
+          case Empty | Leaf(_) | Sequence(_) => Empty
```
jdegoes, line 86: This is a big change in functionality. I think I like it but it does have lots of downstream implications...

core/src/main/scala/zio/config/ConfigDescriptor.scala
afsalthaj: @jdegoes  Could you explain a bit more it?
The default behaviour of just picking up the head ?

core/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -292,10 +296,13 @@ object PropertyTree {
   def unflatten[K, V](map: Map[Vector[K], ::[V]]): List[PropertyTree[K, V]] =
     mergeAll(map.toList.map(tuple => unflatten(tuple._1.toList, tuple._2)))
 
-  def mergeAll[K, V](list: List[PropertyTree[K, V]]): List[PropertyTree[K, V]] =
-    list.foldRight(List[PropertyTree[K, V]](PropertyTree.empty)) {
-      case (tree, acc) => acc.flatMap(tree0 => tree.merge(tree0))
-    }
+  def mergeAll[K, V](list: List[PropertyTree[K, V]]): List[PropertyTree[K, V]] = list.reverse match {
+    case Nil => PropertyTree.empty :: Nil
```
jdegoes, line 300: If we need a guarantee that `mergeAll` produces non empty, we should return `::` here. But I'm not sure we need that guarantee since a user can always do that themselves with the return value of `mergeAll`. So you could return `Nil` here and fix anything if it breaks.

core/src/main/scala/zio/config/ConfigSource.scala
senia-psm: `fromMap` wraps every single value in `Sequence`, so there is no way to parse it as scalar.
There are also some tricky cases with empty `Sequence` in `fromMap` and `fromCommandLineArgs`, though there is no way to describe empty List in `Map[String, String]` or in command line arguments.

core/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -292,10 +296,13 @@ object PropertyTree {
...
+    case head :: tail =>
+      tail.foldLeft(List(head)) {
+        case (acc, tree) => acc.flatMap(tree0 => tree.merge(tree0))
```
jdegoes, line 303: The fact that you rewrote this function indicates to me that `foo merge empty` does yield `foo` anymore. Is that true? Merge with empty should always yield the other side...

core/src/main/scala/zio/config/PropertyTree.scala
senia-psm: Actually no. There is only 1 use case of this method.

core/src/main/scala/zio/config/ReadError.scala
```diff
@@ -37,9 +20,13 @@ sealed trait ReadError[A] extends Exception { self =>
...
 
 object ReadError {
-  final case class MissingValue[A](path: Vector[Either[Int, A]])                     extends ReadError[A]
-  final case class FormatError[A](path: Vector[Either[Int, A]], message: String)     extends ReadError[A]
-  final case class ConversionError[A](path: Vector[Either[Int, A]], message: String) extends ReadError[A]
-  final case class OrErrors[A](list: List[ReadError[A]])                             extends ReadError[A]
-  final case class AndErrors[A](list: List[ReadError[A]])                            extends ReadError[A]
+  sealed trait Step[+K]
+  case class KeyStep[+K](key: K)   extends Step[K]
+  case class IndexStep(index: Int) extends Step[Nothing]
```
jdegoes, line 25: 👍 

Add `final` to both case classes here.

jdegoes: Ah and put `KeyStep` and `IndexStep` into the companion object of `Step`, which will let you shorten `KeyStep` into `Key`, and `IndexStep` into `Index`.

core/src/main/scala/zio/config/PropertyTree.scala
senia-psm: I was trying to avoid sporadic empty nodes.
It's not clear where is the source for those empty nodes. I'll might try to reimplement this functionality in a separate PR.

core/src/main/scala/zio/config/ConfigDescriptor.scala
```diff
@@ -170,15 +171,66 @@ object ConfigDescriptor {
   val float: ConfigDescriptor[String, String, Float] =
     ConfigDescriptor.Source(ConfigSource.empty, PropertyType.FloatType) ?? "value of type float"
 
+  def first[K, V, A](path: K)(desc: ConfigDescriptor[K, V, A]) =
```
jdegoes, line 174: Probably `head` would be more consistent with Scala, since it's the head of a `list`. Also we should have the variant without `path`, like for the other descriptors.

jdegoes, line 174:
```suggestion
  def head[K, V, A](path: K)(desc: ConfigDescriptor[K, V, A]) =
```

core/src/test/scala/zio/config/testsupport/MapConfigTestSupport.scala
```diff
@@ -40,27 +40,30 @@ object MapConfigTestSupport {
 
     object AwsConfig {
       val description: ConfigDescriptor[String, String, AwsConfig] =
-        nested("aws")(
-          (string("key") |@| string("secret") |@| KinesisConfig.description)(AwsConfig.apply, AwsConfig.unapply)
+        first("aws")(
```
jdegoes, line 43: Why is there a need to change all these to `first`? They weren't lists before, right?

core/src/main/scala/zio/config/PropertyTree.scala
jdegoes: 👍

core/src/main/scala/zio/config/ConfigSource.scala
jdegoes: Maybe `fromMap` should be changed so that it only wraps values in Sequence if there is more than one of them?

I agree on the problem of no way to describe empty lists that come in from a map. I'm not sure if we have a good solution to that, other than treating it specially when reading (e.g. allow reading an `Empty` to count as reading an empty list)

core/src/main/scala/zio/config/ConfigDescriptor.scala
jdegoes: I think `Sequence` should be strict like done here, but that the `list` config should be a bit forgiving an accept scalars (as is done in this PR).

core/src/main/scala/zio/config/ConfigDescriptor.scala
jdegoes: Ah, I see. Let's just break backward compatibility here. The new way is much more precise and consistent.

core/src/test/scala/zio/config/testsupport/MapConfigTestSupport.scala
senia-psm: `head("a")(string)` accepts both `a: "v"` `a: ["v1", "v2"]` as input. There is no need for list in description for `head`. There should be a list in configs and this test generates lists in place of scalars.

core/src/main/scala/zio/config/ConfigDescriptor.scala
jdegoes: I think we should not keep backward compatibility, because the old way was not precise.

core/src/main/scala/zio/config/PropertyTree.scala
jdegoes: Yes, but that method is the whole basis of `read`. Anyway, if the tests pass, I'm happy with it!

afsalthaj:
@senia-psm Thanks for your work here. Good one! 
Some of the changes are fixing some real pain (of which empty list is the main one).

The only concern is the use of `first`in an existing test case. I understand why, technically. But we wouldn't need to explain to users about the list semantics in an example where we don't have list.

core/src/main/scala/zio/config/PropertyTree.scala
senia-psm: That the point: `read` is reimplemented.

senia-psm:
> The only concern is the use of `first`in an existing test case. I understand why, technically. But we wouldn't need to explain to users about the list semantics in an example where we don't have list.

My best shot here is a separate syntax objects idea. It's not perfect, but I believe in most cases scalar descriptors should reject list values.

typesafe/src/test/scala/zio/config/typesafe/TypesafeConfigSpec.scala
senia-psm: done

core/src/main/scala/zio/config/ConfigDescriptor.scala
senia-psm: done

core/src/main/scala/zio/config/ConfigDescriptor.scala
senia-psm: done

core/src/main/scala/zio/config/ConfigSource.scala
senia-psm: done with by-name parameters + `lazy val`

core/src/main/scala/zio/config/ConfigSource.scala
senia-psm: done

core/src/main/scala/zio/config/ConfigSource.scala
senia-psm: done

core/src/main/scala/zio/config/PropertyTree.scala
senia-psm: done

core/src/main/scala/zio/config/ReadError.scala
senia-psm: done

core/src/main/scala/zio/config/ConfigDescriptor.scala
senia-psm: done

PR #264, opened 2020-04-07
Fix mixed lists processing

senia-psm:
Allow mixed lists, i.e.:
```
list = [
  "a",
  {b: "c"}
]
```
See updated test.

Also simplify `TypeSafeConfigSource` using the fact, that `ConfigValue` is a "Java-style ADT".


jdegoes:
@senia-psm Thanks for your work on this!

Issue #240, opened 2020-04-01
Support other common data types


jdegoes:
- `UUID`
 - `Duration` / `FiniteDuration` (ZIO)
 - `LocalDate`
 - `LocalDateTime`
 - `LocalTime`
 - `Instant`
 - `URL`
 - `Boolean` (supporting things like `true` / `1` / `on`, and `false` / `0` / `off`)
 - `File`
 - `Path`
 - `BigInteger`
 - `BigDecimal`
 - All primitives (`Short`, `Byte`, etc.)
 - Tuples
 - Sets

Any others?

marekklis:
PR for `UUID` https://github.com/zio/zio-config/pull/268

marekklis:
I'll work on `Set` support

afsalthaj:
This is done.

PR #272, opened 2020-04-08
Support File for PropertyType in  #240

cecol:
For #240, if any problem or bug, please tell me.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -14,6 +14,6 @@ optIn.annotationNewlines = true
 
 rewrite.rules = [SortImports, RedundantBraces]
 project.excludeFilters = [
-  "zio-config/src/main/scala/zio/config/ProductBuilder.scala",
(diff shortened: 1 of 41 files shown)


afsalthaj:
Hey could you do sbt fmt and push again ?

cecol:
Sorry, I forget to adjust my Intellij IDEA format preference. I fix it and pass `sbt check` on my local PC.

jdegoes:
@cecol Can you just fix conflicts and then @afsalthaj can merge?

afsalthaj:
@cecol should we actually check for existence for File ? I guess we can avoid that step?

cecol:
@afsalthaj My original code didn't check the existence of the File, but that would get another problem of PropertyTypeTest. I have no idea how to program the invalid test case. since the `new File(_)`  accepts any arbitrary String without exceptions.

PR #291, opened 2020-04-11
fix #271 add links to README header

### LINKED ISSUE #271: Create maven version thumbnail in README

senia-psm:
shamelessly stolen from zio/zio


afsalthaj:
Well. We should be copying things from zio whenever it’s possible. Good one!

PR #286, opened 2020-04-10
Implement map and revisit write to hocon functionality + incremental code clean up

afsalthaj:

PR #292, opened 2020-04-11
Fix write functions

afsalthaj:
Tests for 
```scala
read(descriptor from typesafeSource) == read(descriptor from write(descriptor, read(descriptor from typesafeSource)))

```
And all the fixes to make it work!

To be reviewed once #286  is merged


afsalthaj:
Once #286 is merged, we can see the difference

afsalthaj:
@senia-psm I am pushing the very latest change. There were a few more bugs

afsalthaj:
You might need to approve again

typesafe/src/main/scala/zio/config/typesafe/TreeToHoconSupport.scala
```diff
@@ -17,7 +17,11 @@ private[typesafe] trait PropertyTreeFunctions {
     tree match {
       case leaf @ Leaf(_)     => loopLeaf(key, leaf)
       case record @ Record(_) => loopRecord(record)
-      case PropertyTree.Empty => ConfigFactory.empty().root()
+      case PropertyTree.Empty =>
+        key.fold(ConfigFactory.empty().root()) { k =>
+          ConfigFactory.empty().withValue(k, ConfigValueFactory.fromAnyRef(null: Option[String])).root()
+        }
```
afsalthaj, line 23: Ofcourse, if `empty` it is not `x : {}` which corresponds to ` ConfigFactory.empty().root()`.
This was breaking before since we now pattern match on the type of tree we looking at, and it always gave `Record`

core/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -164,7 +132,9 @@ sealed trait PropertyTree[+K, +V] { self =>
               }
           }
           .map(v => PropertyTree.Record(v))
-      case (l, r) => l :: r :: Nil
+      case (left, right) if left.isEmpty  => singleton(right)
+      case (left, right) if right.isEmpty => singleton(left)
+      case (l, r)                         => l :: r :: Nil
```
afsalthaj, line 137: This was important, since it skipped the right side when empty

Issue #223, opened 2020-03-14
Pretty print the errors (may be) similar to zio.Cause


afsalthaj:
Prettyprint ReadErrors

cecol:
I am working on this. Hope this can be done this weekend

PR #297, opened 2020-04-12
Add ReadError prettyPrint and fix some typo

cecol:
For issue #223, I just simply take the idea from `zio.Cause.prettyPrint`.
Any idea is preferable


afsalthaj:
Thanks for the initiative. Yes, we did think zio.Cause as an inspiration and stay consistent. Mind giving an example of how prettyPrint actually looks like?

afsalthaj:
We can have multiple prettyPrint if that’s not too difficult. The above one is not too bad. We Will have json/hocon as well. 
 
or : [ 
    error1,
    and : [ 
        error2,
        or : [
           error3,
           error4
         ]
     ]
]

cecol:
Below configuration
```scala
val databaseConfig =
    (string("url") |@| int("port"))(Database.apply, Database.unapply)

  val configuration =
    (int("port") |@| string("addr").orElse(string("addr2")) |@| nested("db")(databaseConfig))(
      AppConfig.apply,
      AppConfig.unapply
    )
```
would generate
```shell
ReadError failed.
╥
╠══╦══╦══╦══╗
║  ║  ║  ║  ║
║  ║  ║  ║  ╠─FormatError
║  ║  ║  ║  ║ Provided value is fakePort, expecting the type int
║  ║  ║  ║  ║ Key(db)
║  ║  ║  ║  ║ Key(port)
║  ║  ║  ║  ▼
║  ║  ║  ║
║  ║  ║  ╠─FormatError
║  ║  ║  ║ Provided value is of type Sequence, expecting the type Leaf
║  ║  ║  ║ Key(db)
║  ║  ║  ║ Key(url)
║  ║  ║  ▼
║  ║  ║
║  ║  ╠─MissingValue
║  ║  ║ Key(addr)
║  ║  ║
║  ║  ╠─MissingValue
║  ║  ║ Key(addr2)
║  ║  ▼
║  ║
║  ╠─FormatError
║  ║ Provided value is 123xyz, expecting the type int
║  ║ Key(port)
║  ▼
```

The original ReadError `AndErrors(List(AndErrors(List(FormatError(List(Key(port)), Provided value is 123xyz, expecting the type int), OrErrors(List(MissingValue(List(Key(addr))), MissingValue(List(Key(addr2)))))), AndErrors(List(FormatError(List(Key(db), Key(url)), Provided value is of type Sequence, expecting the type Leaf), FormatError(List(Key(db), Key(port)), Provided value is fakePort, expecting the type int)))`

afsalthaj:
║  ║  ║  ║ Key(db)
║  ║  ║  ║ Key(url)
║  ║  ║  ▼

It could be Key(db.url)
Such that it scales for too many level of nesting

cecol:
> ║ ║ ║ ║ Key(db)
> ║ ║ ║ ║ Key(url)
> ║ ║ ║ ▼
> 
> It could be Key(db.url)
> Such that it scales for too many level of nesting

yeah, it is better. Let me work on this

cecol:
new look
```text
ReadError failed.
ReadError failed.
╥
╠══╦══╦══╦══╗
║  ║  ║  ║  ║
║  ║  ║  ║  ╠─FormatError
║  ║  ║  ║  ║ cause: Provided value is fakePort, expecting the type int
║  ║  ║  ║  ║ path: db.port
║  ║  ║  ║  ▼
║  ║  ║  ║
║  ║  ║  ╠─FormatError
║  ║  ║  ║ cause: Provided value is of type Sequence, expecting the type Leaf
║  ║  ║  ║ path: db.url
║  ║  ║  ▼
║  ║  ║
║  ║  ╠─MissingValue
║  ║  ║ path: addr
║  ║  ║
║  ║  ╠─MissingValue
║  ║  ║ path: addr2
║  ║  ▼
║  ║
║  ╠─MissingValue
║  ║ path: portNowhere
║  ║
║  ╠─FormatError
║  ║ cause: Provided value is 123xyz, expecting the type int
║  ║ path: port
║  ▼
▼
```

senia-psm:
@cecol Please add a test with full text for several errors.

cecol:
> @cecol Please add a test with full text for several errors.

I already added it, please check. Thanks

core/src/test/scala/zio/config/ReadErrorsTest.scala
```diff
@@ -36,4 +41,82 @@ object ReadErrorsTestUtils {
...
+      |║  ║  ║  ║  ║  ║  ╠─FormatError
+      |║  ║  ║  ║  ║  ║  ║ cause: Format error
+      |║  ║  ║  ║  ║  ║  ║ path: k1.k2
+      |║  ║  ║  ║  ║  ║  ▼
+      |║  ║  ║  ║  ║  ║
+      |║  ║  ║  ║  ║  ╠─MissingValue
+      |║  ║  ║  ║  ║  ║ path: k1.k2.1.k3
+      |║  ║  ║  ║  ║  ║
+      |║  ║  ║  ║  ║  ╠─MissingValue
+      |║  ║  ║  ║  ║  ║ path: k1.k2.1.k3
```
senia-psm, line 80: I'd suggest to print index this way: `k1.k2[1].k3`

core/src/test/scala/zio/config/ReadErrorsTest.scala
```diff
@@ -36,4 +41,82 @@ object ReadErrorsTestUtils {
...
+        )
+      ),
+      ReadError.AndErrors(List(ReadError.OrErrors(List(simpleMissingValue, simpleMissingValue)), simpleFormatError))
+    )
+  )
+
+  val complexErrorsPrettyPrint =
+    """ReadError failed.
+      |╥
+      |╠══╦══╦══╦══╦══╦══╗
```
senia-psm, line 69: Is there a limit on widht? There might be a lot more errors.

senia-psm: For instance: in ADT you'll get N branches for N subclasses.

senia-psm:
@cecol it looks really cool and quite readable.
So we have `OrError`s as parallel lines and `AndErrors` on the same line? That's a brilliant idea!

core/src/test/scala/zio/config/ReadErrorsTest.scala
cecol: There is no limit on width. The parallel lines number is same as the total errors in AndErros.

afsalthaj:
@senia-psm its AndErrors in parallel and OrErrorz in same line. 
@cecol isn’t it? 
What’s our take on this? Should we try the other way around?

afsalthaj:
Anyway in this way, we are consistent with zio.Cause errors and that’s really a good win. Good work on this

afsalthaj:
If parallel lines is and-errors, then given the fact that, in common usecases, the number of and-errors will be greater than number of or-errors resulting in more number of lines resulting in cluttered situation once it gone past the width of console/terminal. On the other hand, if and-errors are same line, it will seem a bit cluttered that everything in one single line. We need to choose which one will act least worse amongst these 2 options in the given scenario.

cecol:
> @senia-psm its AndErrors in parallel and OrErrorz in same line.
> @cecol isn’t it?
> What’s our take on this? Should we try the other way around?

Yes, only the AndErrors creates new lines for other errors and OrErrorz stays the same line
That is this ReadError prettyPrint core idea

Issue #293, opened 2020-04-11
Add error accumulation tests


afsalthaj:
This should be fairly straight forward. There are already error accumulation tests for simple types. Extend them for maps and lists (which is being added recent days)

afsalthaj:
This should be fairly straight forward. There are already error accumulation tests for simple types. Extend them for maps and lists (which is being added recent days)

PR #300, opened 2020-04-14
Errors accumulation tests for list, set and map

marekklis:
For https://github.com/zio/zio-config/issues/293


core/src/test/scala/zio/config/SetTest.scala
```diff
@@ -271,6 +271,36 @@ object SetTest
...
+                AndErrors(
+                  List(
+                    FormatError(List(Key("a"), Index(1)), "Provided value is lorem ipsum, expecting the type boolean")
+                  )
+                ),
+                AndErrors(List(FormatError(List(Key("b"), Index(0)), "Provided value is one, expecting the type int")))
+              )
+            )
+
+          assert(res)(isLeft(equalTo(expected)))
```
afsalthaj, line 303: Let's test whether index associated with errors test separately.
For list and map

Also, great to have another complicated case class test with all of them failed.

afsalthaj: Essentially the idea is, this test should also act as a good README to understanding the error messages.
Also now you can have one more test that gives out the pretty print version (may be, or may not be relevant)

core/src/test/scala/zio/config/SetTest.scala
```diff
@@ -271,6 +271,113 @@ object SetTest
...
+                  )
+                ),
+                AndErrors(
+                  List(
+                    FormatError(List(Key("b"), Index(1)), "Provided value is one, expecting the type int")
+                  )
+                )
+              )
+            )
+          assert(res)(isLeft(hasField("size", _.size, equalTo(expected.size))))
```
marekklis, line 328: I think that `ConversionError(List(Key("b"), Index(0)), "Duplicated values found")` error is missing for that case. Am I right?

core/src/test/scala/zio/config/SetTest.scala
marekklis: > Let's test whether index associated with errors separately.

how can I do that? 
I found only 2 possibilities to assert that case:
```
assert(res)(isLeft(equalTo(expected)))
```
or 
```
assert(res)(isLeft(hasField("size", _.size, equalTo(expected.size))))
```

core/src/test/scala/zio/config/SetTest.scala
afsalthaj: @marekklis  I am not sure. Is it a bug you talking about?

marekklis: @afsalthaj yes, I think this is a bug. If `b` is a set of int, for the following input:
```
"b" -> Sequence(Leaf("2") :: Leaf("one") :: Leaf("2") :: Nil)
```
we should get two errors accumulated:
`FormatError(..., "Provided value is one, expecting the type int")` and `ConversionError(..., "Duplicated values found")` but we there is only `FormatError` returned.

afsalthaj: @marekklis Sorry for late response. Sure, is this only in the case of Set that we have this corner case? I guess the issue is it is not able to parse all the individual elements first to then “Try” forming a set. I think, it’s a bug, but it is ok to merge this current PR once you fix the build issue.
Let’s keep track of the expectation with set as a separate issue.

marekklis:
@afsalthaj Compilation errors fixed. Could you take a look once again because your approval was dismissed by my last commit

PR #303, opened 2020-04-15
Add narrow syntax helper to zoom into configs

vpavkin:
### Motivation

There's no easy way to grab a subset of config to pass it down to a service, that needs only that subset. It is a drag in real-world apps, making separation of concerns painful. [Detailed case on discord](https://discordapp.com/channels/629491597070827530/633028431000502273/689138668392415280).

### Proposal

This PR adds a simple syntax extension to `ZLayer`: with it you can create layers for nested configurations right in place, like this:

```scala
import zio.config.syntax._ // extension is here

val cfg = TypesafeConfig.fromHoconString(str, AppConfig.description)
cfg.narrow(_.api) >>> endpoint
cfg.narrow(_.db) >>> repository
```

It uses `ZLayer` map which means no additional side-effects are executed in these child layers, as demonstrated by [this example](https://scastie.scala-lang.org/svgR5W67Q2CCOgIRA2WU1Q).

**P. S.** Went with pretty simple `syntax` package to put the extension to. Open to suggestions to improve it.

### DIFF
--- a/core/src/main/scala/zio/config/syntax/package.scala
+++ b/core/src/main/scala/zio/config/syntax/package.scala
@@ -1,0 +1,11 @@
+package zio.config
+
+import zio.{ Has, Tagged, ZLayer }
+
+package object syntax {
+
+  final implicit class ZIOConfigNarrowOps[R, E, A](val self: ZLayer[R, E, Has[A]]) extends AnyVal {
+    def narrow[B: Tagged](f: A => B)(implicit T: Tagged[A]): ZLayer[R, E, Has[B]] =
+      self.map(a => Has(f(a.get)))
+  }
+}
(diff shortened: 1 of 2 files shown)


vpavkin:
btw, @afsalthaj, do you want me to add a note about this feature somewhere in the docs?
If so, a pointer of where it's best to be done would be great :)

jdegoes:
@vpavkin You could submit this to ZIO if you like. 😄

afsalthaj:
@vpavkin As Johb mentioned It's a good idea to put that into ZIO. May be we can merge this one, and let devs start using it and by the time we have the functionality in ZIO 1.0, we can delete it from here.
On docs, I think , its more of a zio thing. Let's have that in zio docs.

Issue #266, opened 2020-04-08
Document command line args


afsalthaj:
Document command-line-args based on the below examples in website:
https://github.com/zio/zio-config/tree/master/examples/src/main/scala/zio/config/examples/commandline

PR #305, opened 2020-04-17
Add commandline doc

cecol:
for #266


docs/sources/index.md
```diff
@@ -294,4 +294,37 @@ TypesafeConfig.fromHoconString(jsonString, automaticDescription)
...
+```
+The config with a complex source of command line arguments is what every production service needs
+```scala mdoc:silent
+final case class UserPassword(k2: String, k3: String)
+final case class DatabaseConfig(conf: UserPassword, url: String)
+final case class VaultConfig(userPassword: UserPassword)
+final case class SparkConfig(databaseConfig: DatabaseConfig, numberOfExecutors: Int)
+final case class AppConfig(sparkConfig: SparkConfig, vault: VaultConfig, users: List[String], region: List[String])
+
+val argss = "--conf -database.username=1 --conf -database.password=hi --conf.database.url=jdbc://xyz --conf -num_execs=10 --vault.username=3 --vault.password=10 --vault.something=11 --users 1 --users 2 --region 111,112"
```
afsalthaj: In this example let’s talk about only one type of nesting and that is nesting using dots. 

That’s is talk about only —conf.database.url.....  
and not —conf database.url

This will create less confusions amongst users. 


We can add another example that talks about “hey you could also do —conf -num_executors=1   
—conf -num_tasks=2 —conf -num_mappers=10 , where you can consider conf as a way to nest. 
for example, either retrieve the map using map(conf)(string) retrieving all the key value pairs under spark ”conf”. Or even similar case classes as above. 

As a third example, we can talk about lists

—url bla.com —url xyz.com —url abc.com

list(url)(string) returns all the urls 

This is really useful for many users. Especially, it is useful when users really run of an actual delimiter because their url is complex enough with :: and ; and ,  etc.

And lets mention we can also do 

—url bla.com,xyz.com,abc.com 
And pass a value delimiter to get them as list using the same description.

afsalthaj: We can restrict some functionalities depending on the feedback. As of now, let’s don’t mix all the patterns in 1 example. Instead, as I suggested in the above example, we can have multiple small examples

afsalthaj: Also I think we did this command line a day. We can give more thoughts on to this going forward. Let’s  say “experimental” in the document similar to what we did with hocon in the beginning.  


Also let’s explicitly mention x=y doesn’t work because x doesn’t start with - . I have seen both patterns, so to be honest I don’t know if it’s right or wrong to have that restriction.

afsalthaj:
Very useful PRs. We should have more PRs into docs similar to this one.

afsalthaj:
Also make sure we have the correct code base in Mdoc. Mdoc runs the code snippets and if it throws exception (runtime) build fails.

cecol:
> Also make sure we have the correct code base in Mdoc. Mdoc runs the code snippets and if it throws exception (runtime) build fails.

Updated
Mdoc is so helpful!! I had no idea about Modc, and learn a lot from this project.

afsalthaj:
Sure let’s add the examples I mentioned and all good to merge straight away

cecol:
> Sure let’s add the examples I mentioned and all good to merge straight away

I updated the nested case and mention the `—url bla.com,xyz.com,abc.com` case.
If any misunderstanding, I would like to improve them.

docs/sources/index.md
```diff
@@ -293,5 +292,44 @@ TypesafeConfig.fromHoconString(jsonString, automaticDescription)
 
 
 ```
-
 Please check other ways to load the hocon file in `TypesafeConfig`
+
+## Command Line Arguments Source
+Simply source config from command line arguments
```
afsalthaj, line 298: Make the heading `Simple command line arguments` a subheading. That is by adding `###`

`### Simple command line arguments`

docs/sources/index.md
```diff
@@ -293,5 +292,44 @@ TypesafeConfig.fromHoconString(jsonString, automaticDescription)
...
+```scala mdoc:silent
+final case class SimpleCommandLineConfig(key1: String, key2: String)
+
+val simpleCmdLineArgs = "--key1 value1 --key2 value2"
+val simpleSource = ConfigSource.fromCommandLineArgs(simpleCmdLineArgs.split(' ').toList)
+val simpleConfig = descriptor[SimpleCommandLineConfig] from simpleSource
+```
+With nested path in command line arguments
```
afsalthaj, line 306: Remove line number 306, and add the below heading - its a subheading, meaning , `###`

### Nested config using command line arguments: Approach 1

```
 final case class SparkConf(key1: String, key2: String)
 final case class NestedCommandLineConfig(conf: SparkConf, key3: String)

 val nestedCmdLineArgs = "--conf.key1 v1 --conf.key2 v2 --key3 v3"
 
val nestedSource = 
   ConfigSource.fromCommandLineArgs(
     nestedCmdLineArgs.split(' ').toList, 
     keyDelimiter = Some('.')
  )

 val nestedConfig = descriptor[NestedCommandLineConfig] from nestedSource
 assert (read(nestedConfig)  == Right(<your result here>)

```
This config is for those dev who really used to system properties (-Dconf.key=1) and want to take the same approach towards command line arguments. Here we make use of delimiter `.` as the tool to nesting.

For those who hate delimited keys in command-line and the associated nesting, we will have different approach as given below

### Nested config using command line arguments: Approach 2

```
 val nestedCmdLineArgs2 = "--conf -key1=v1 --conf -key2=v2 --key3 v3"

 val nestedSource2 = 
  ConfigSource.fromCommandLineArgs(nestedCmdLineArgs.split(' ').toList)

 val nestedConfig2 = descriptor[NestedCommandLineConfig] from nestedSource2

 assert (read(nestedConfig2)  == Right(<your result here>)

```
Here we don't use delimiters for nesting, hence `keyDelimiter` is None. In this case any key value that comes after `--conf` comes under the root path `conf`. This is followed in various places such as spark.

In fact, we can go any level nesting. For example,  we can give  `---aws --db -url="v" ---aws --kinesis -topic=x`, although let's don't complicate our command line arguments.

### Map using command line arguments
Both the approaches that we saw with `nesting` is applicable to map.

```scala
val args = --conf.key1=value1  --conf.key2=value2

`map("conf")(string)` retrieving `Map("key1" -> "value1", "key2" -> "value2")`

```
This will also work if args is `--conf -key1=value1 --conf -key2=value2`

docs/sources/index.md
```diff
@@ -293,5 +292,44 @@ TypesafeConfig.fromHoconString(jsonString, automaticDescription)
...
+```scala mdoc:silent
+final case class SparkConf(key1: String, key2: String)
+final case class NestedCommandLineConfig(conf: SparkConf, key3: String)
+
+val nestedCmdLineArgs = "--conf.key1 v1 --conf.key2 v2 --key3 v3"
+val nestedSource = ConfigSource.fromCommandLineArgs(nestedCmdLineArgs.split(' ').toList, keyDelimiter = Some('.'))
+val nestedConfig = descriptor[NestedCommandLineConfig] from nestedSource
+```
+The config with a complex source of command line arguments is what every production service needs
```
afsalthaj, line 315: Remove line number 315. Add the below heading. The heading is a subheading which is `###`
### Lists in Command Line Arguments: Approach 1

```
val listArgs = "--users Jane --users Jack"
val listSource = ConfigSource.fromCommandLineArgs(listArgs2.split(' ').toList
val config = list("users")(string) from listSource
assert( 
   read(config) == Right("Jane", "Jack")
)

```

afsalthaj: Add another heading
### Lists in Command Line Arguments (alternate approach)

```
val listArgs2 = "--users Jane,Jack"

// args.split(' ') is only for demo purpose. We already get a list if we use zio.App
val listSource2 = ConfigSource.fromCommandLineArgs(
   listArgs2.split(' ').toList,
   valueDelimiter = Some(',')
 )

list("user")(string) from listSource2 
```

docs/sources/index.md
```diff
@@ -293,5 +292,44 @@ TypesafeConfig.fromHoconString(jsonString, automaticDescription)
...
+```scala mdoc:silent
```
afsalthaj, line 316: Heading here 

### A complex production application config (demo)

docs/sources/index.md
```diff
@@ -293,5 +292,44 @@ TypesafeConfig.fromHoconString(jsonString, automaticDescription)
...
+  complexArgs.split(' ').toList,
+  Some('.'),
+  Some(',')
+)
+val appConfig = descriptor[AppConfig] from complexSource
+```
+* From above, command line arguments `--users 1 --users 2` is the source of `users: List[String]` config.
+    * A ConfigDescriptor `list(url)(string)` returns all the urls with command line arguments: `—url bla.com —url xyz.com —url abc.com`  
+* With an assisted delimiter `','`, you can source `region: List[String]` from command line arguments: `--region 111,112`.
+* For nested path, `val args = "—conf -database.username=1"` can be another way to nest, and using `map(conf)(string)` can retrieve all the key-value pairs under `conf: SparkConfig`.
```
afsalthaj, line 335: Since we already explained about list, map etc before, we don't need much explanation here.
Remove line number 332 to 335

afsalthaj:
@cecol I have added the additional headings that we need, and the gist of the content.
Please make them as consistent as possible. Thanks

docs/sources/index.md
afsalthaj: Please keep them as consistent as possible. I hope you find a pattern here

afsalthaj:
@cecol  In the process of updating the docs, if you find any inconsistent in existing docs, please feel free to fix it.

Inconsistencies can be :

- Unnecessary uppercase in various places 
- Inconsistent heading pattern (Use of ## and ### etc)
- Verbose explanation demotivating users to read.
- Missing space after `.`. 
- Missing uppercase after `.` 
- Spelling errors (if any)

docs/sources/index.md
cecol: I initially had that intent, but not confident so much. Let's make it

cecol:
@afsalthaj Thanks for your explanation, I updated them all

PR #307, opened 2020-04-17
Fix typo in name of an object

whysoserious:
I am not sure if this is a correct change but there you go.


afsalthaj:
Thanks @whysoserious !

PR #308, opened 2020-04-18
Modularising ZIO-Config and reduce leaking out type parameters

afsalthaj:
General improvements and bug fixes

The major improvement in this PR is modularising and stop leaking `K` and `V` type parameters to users.

With backward compatibility and keeping the same level of polymorphism, we manage to convert  `ConfigDescriptor[K, V, A]` to a friendly `ConfigDescriptor[A]`.

Most of the files are modules now.

During the process, ProductBuilder becomes abstract and independent of configdescriptor, and now consist of an abstract zip and xmapEither. The complexity is anyway not exposed to user

https://github.com/zio/zio-config/blob/module/core/src/main/scala/zio/config/ProductBuilder.scala

```scala
private[config] trait ProductBuilder[F[_], A, B] { self =>

  def zip[X, Y]: (F[X], F[Y]) => F[(X, Y)]
  def xmapEither[X, Y]: (F[X], X => Either[String, Y], Y => Either[String, X]) => F[Y]

  implicit class Syntax[X](x: F[X]) {
    def zip[Y](y: F[Y]): F[(X, Y)] = self.zip(x, y)
    def xmapEither[Y](f: X => Either[String, Y], g: Y => Either[String, X]): F[Y] = self.xmapEither(x, f, g)
  }

  val a: F[A]
  val b: F[B]

```


core/src/main/scala/zio/config/Config.scala
```diff
@@ -6,6 +6,7 @@ import zio.system.System
 import zio.{ Layer, Tagged, ZIO, ZLayer }
 
 object Config {
+  import string._
```
afsalthaj, line 9: May be we can pull this off to `object string`

afsalthaj:
Looking for early suggestions before fixing the enormous number of errors in Scala 2.12

core/src/main/scala/zio/config/string.scala
```diff
@@ -0,0 +1,114 @@
+package zio.config
+
+import java.io.File
+import java.net.{ URI, URL }
+import java.time.{ Instant, LocalDate, LocalDateTime, LocalTime }
+import java.util.UUID
+
+import scala.concurrent.duration.Duration
+
+object string extends ReadFunctions with ConfigDocsFunctions {
```
senia-psm, line 10: I'd suggest to name it not after `K` and `V` types (implementation details), but after its role like `default` or `common`.
And `string.string` looks a little bit misleading.

core/src/main/scala/zio/config/string.scala
```diff
@@ -0,0 +1,114 @@
...
+  type K = String
+  type V = String
+
+  val bigDecimal: ConfigDescriptor[BigDecimal] =
+    Source(ConfigSource.empty, PropertyType.BigDecimalType) ?? "value of type bigdecimal"
```
senia-psm, line 15: `PropertyType` is `V` type based and it's sill on top level without module.

contributing.md
```diff
@@ -0,0 +1,40 @@
...
+1. Please raise an issue in [here](https://github.com/zio/zio-config/issues) before you raise a pull request directly.
+2. Tag the issue in PR description.
+3. Contributions to documentation/website is always welcome and can be raised without tagging an issue.
+
+## SBT
+
+```scala
+cd zio-config
+sbt +compile
+sbt runAllExamples
```
senia-psm, line 13: `PropertiesFileExample.scala` is broken.

afsalthaj: Sure. I will fix it

core/src/main/scala/zio/config/string.scala
afsalthaj: `string` implies a `key` (at any level) is always `string` and the value associated with it will also be `string`.  But I like the name `common`. `default` in `config` can be misleading thinking we are dealing with `default values

core/src/main/scala/zio/config/string.scala
afsalthaj: PropertyType. BigDecimalType can convert the value which can only a string.
I think the instances of `propertytype` should be brought back to `string` (or `default`) module

afsalthaj:
More updates and improvements to follow here

contributing.md
afsalthaj: @senia-psm  This was fixed ok

PR #323, opened 2020-04-29
partially fixes #322 - handle '=' in CLI argument's value

### LINKED ISSUE #322: Handle `=` in values of command line arguments

marioosh:
I've managed this partially only... 

assume cases:
`--key=valuewith=inside` or `--key valuewith=inside`

It's able to easy distinguish value when both key and value are separated with `=` sign. But in case where both are separated with space char it's rather impossible to guess whether `valuewith=inside` is entire value or it's key `valuewith` with value of `inside`.


afsalthaj:
Thanks @marioosh

Issue #325, opened 2020-05-08
[HOCON] TypesafeConfigSource#fromHoconString does not catch parse error


fsarradin:
Hi,

I have got an exception and not an Either instance if i do the following:

```scala
import zio.config.typesafe._

TypesafeConfigSource.fromHoconString("{")
```

This produces the following exception:

```
com.typesafe.config.ConfigException$Parse: String: 1: expecting a close brace or a field name here, got end of file
at com.typesafe.config.impl.Parser$ParseContext.parseError(Parser.java:430)
at com.typesafe.config.impl.Parser$ParseContext.parseError(Parser.java:426)
at com.typesafe.config.impl.Parser$ParseContext.parseKey(Parser.java:583)
at com.typesafe.config.impl.Parser$ParseContext.parseObject(Parser.java:740)
at com.typesafe.config.impl.Parser$ParseContext.parseValue(Parser.java:516)
at com.typesafe.config.impl.Parser$ParseContext.parse(Parser.java:941)
at com.typesafe.config.impl.Parser.parse(Parser.java:34)
at com.typesafe.config.impl.Parseable.rawParseValue(Parseable.java:222)
at com.typesafe.config.impl.Parseable.rawParseValue(Parseable.java:210)
at com.typesafe.config.impl.Parseable.parseValue(Parseable.java:176)
at com.typesafe.config.impl.Parseable.parseValue(Parseable.java:170)
at com.typesafe.config.impl.Parseable.parse(Parseable.java:227)
at com.typesafe.config.ConfigFactory.parseString(ConfigFactory.java:791)
at com.typesafe.config.ConfigFactory.parseString(ConfigFactory.java:795)
at zio.config.typesafe.TypesafeConfigSource$.fromHoconString(TypesafeConfigSource.scala:32)
```

This seems to happen because [`fromHoconString`](https://github.com/zio/zio-config/blob/master/typesafe/src/main/scala/zio/config/typesafe/TypesafeConfigSource.scala#L32) is calling `ConfigFactory.parseString(input).resolve` first, before calling `fromTypesafeConfig`.

AFAIK, a way to solve this issue would be to use a call-by-name instead of a call-by-value. What do you think?

regards

fsarradin:
FYI

```sbt
libVersion.zioConfig = "1.0.0-RC17"
scalaVersion = "2.13.1"
sbt.version = "1.3.9"

  "dev.zio" %% "zio-config"          % libVersion.zioConfig,
[...]

PR #326, opened 2020-05-08
fix fromHoconString producing exception on parse error (#325)

fsarradin:
I have added a call-by-name for `fromTypesafeConfig` and a unit test to check that we get a Left value instead of an exception.

See issue #325

### DIFF
--- a/typesafe/src/test/scala/zio/config/typesafe/TypesafeConfigErrorsSpec.scala
+++ b/typesafe/src/test/scala/zio/config/typesafe/TypesafeConfigErrorsSpec.scala
@@ -101,6 +101,17 @@ object TypesafeConfigErrorsSpec extends DefaultRunnableSpec {
       assert(nestedConfigAutomaticResult3)(equalTo(nestedConfigAutomaticExpect3)) &&
       assert(nestedConfigAutomaticResult4)(equalTo(nestedConfigAutomaticExpect4))
     },
+    test("A variant error case with a not well-formed typesafe HOCON config") {
+      val hocconStringWithParseError =
+        s"""
+         account {
+        """
+
+      val notWellFormedConfigResult = TypesafeConfigSource.fromHoconString(hocconStringWithParseError)
+      val notWellFormedConfigExpect = Left("String: 3: expecting a close parentheses ')' here, not: end of file")
+
+      assert(notWellFormedConfigResult)(equalTo(notWellFormedConfigExpect))
+    },
     test("A variant error case with typesafe HOCON config and a manual description") {
       val configNestedManual = {
         val accountConfig =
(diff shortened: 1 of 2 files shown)


typesafe/src/test/scala/zio/config/typesafe/TypesafeConfigErrorsSpec.scala
```diff
@@ -101,6 +101,17 @@ object TypesafeConfigErrorsSpec extends DefaultRunnableSpec {
...
+    test("A variant error case with a not well-formed typesafe HOCON config") {
+      val hocconStringWithParseError =
+        s"""
+         account {
+        """
+
+      val notWellFormedConfigResult = TypesafeConfigSource.fromHoconString(hocconStringWithParseError)
+      val notWellFormedConfigExpect = Left("String: 3: expecting a close parentheses ')' here, not: end of file")
+
+      assert(notWellFormedConfigResult)(equalTo(notWellFormedConfigExpect))
```
ahoy-jon, line 113: This test is very specific. I would go for something like:
```scala
assert(notWellFormedConfigResult.isLeft)(Assertion.isTrue)
```

fsarradin: Agreed. Modification done.

ahoy-jon:
Indeed we* are using this kind of technique when we are wrapping impure type: 
https://github.com/univalence/zio-spark/blob/master/src/main/scala/zio/spark/wrap/Wrap.scala#L12-L44
(With a lot of dependent typing here)

However, the technique of creation would be something like :
````
object ZTypesafeConfig {
   def fromString(string: String): Task[ZTypesafeConfig]  =  ??? 
   def default: UIO[ZTypesafeConfig]  =   ???
   // In future lot many others may be

}
```
Because the config relies on mutable elements in the JVM environment.

* Working in the same organization, @univalence

ahoy-jon:
@afsalthaj That being said, I don't understand why this PR is not merged, and I see a pattern emerging in ZIO contributions (we (try to) contribute to ZIO).

1. There is an issue with a **harmful** bug, related to the initial design
2. There is a minimal Pull Request
3. The Pull Request comes with a valid test

Yet it is not merged, because the PR is turned as a vehicle to change or to discuss the initial design (with a new contributor).

Could you: 
- create a new issue to change the design?
- merge this PR?

I am not trying to be mean, just exaggerating a bit! 😀

afsalthaj:
The change I proposed doesn’t need any redesign. It doesn’t need to return a `Task` in smart constructor space.

It’s a small change. Merging this one, once the build is over.

afsalthaj:
@ahoy-jon  Thanks for your contribution

PR #327, opened 2020-05-10
Add unit test for simple list for fromMap

leigh-perry:


core/src/test/scala/zio/config/MapTest.scala
```diff
@@ -57,6 +57,27 @@ object MapTest
 
           assert(res)(isRight(equalTo(Cfg("sa", Map("z" -> List("d"))))))
         },
+        test("list of delimited values from Map.") {
+          case class Cfg(a: String, b: List[String])
+
+          val cCfg = (string("a") |@| list("b")(string))(Cfg, Cfg.unapply)
```
afsalthaj, line 63: Let’s rename `cCfg` to `config` ?

PR #343, opened 2020-05-27
Fix #338

### LINKED ISSUE #338: optional list does not error if items in the list are not correct
zio-config version: 1.0.0-RC18
scala 2.12

**Expected behaviour:**
Given `Option[List[Branch]]`, if one of the `Branch`s is not defined correctly (`pattern` attribute is wrongly named below as `name`) then parsing results in a failure.

**Actual behaviour:**
Result is `None`

Repro here: https://scastie.scala-lang.org/ebscQhvmS7mgChR5agRwgw
or
```
import scala.language.postfixOps

case class AppConfig(branches: Option[List[Branch]])
case class Branch(pattern: String, tag: Boolean)
val stringg = 

"""
branches: [
  {
    pattern: "master"
    tag: true
  }
  {
    name: "pull/.*"
    tag: true
  }
]
"""

import zio.config.typesafe._
import zio.config._, ConfigDescriptor._

  val patternDesc = string("pattern").describe("regex to match branch name on")
  val tagDesc = boolean("tag").describe("whether to actually tag this branch when `mkver tag` is called")

  val branchConfigDesc = 
    (
      patternDesc |@|
      tagDesc
    )(Branch.apply, Branch.unapply)

  val appConfigDesc = 
    (listStrict("branches")(branchConfigDesc).optional)(AppConfig.apply, AppConfig.unapply)



val source = TypesafeConfigSource.fromHoconString(stringg) match {
  case Right(a) => a
  case Left(_) => throw new Exception("bad hocon string")
}

// Expect this to fail becuase branches[1].name is incorrect.
// It should be branches[1].pattern according to config desc
read(appConfigDesc from source).swap.map(_.prettyPrint()).swap
```

Dmitrii-Shevchenko:
Add ForceSeverity error
Add tests with unrecoverable errors on map and list

### DIFF
--- a/core/src/main/scala/zio/config/ReadError.scala
+++ b/core/src/main/scala/zio/config/ReadError.scala
@@ -3,20 +3,22 @@ package zio.config
 import ReadError._
 sealed trait ReadError[A] extends Exception { self =>
   override def toString: String = self match {
-    case ReadError.MissingValue(path)             => s"MissingValue(${path})"
-    case ReadError.FormatError(path, message)     => s"FormatError(${path}, ${message})"
-    case ReadError.ConversionError(path, message) => s"ConversionError(${path},${message}"
-    case ReadError.OrErrors(list)                 => s"OrErrors(${list.map(_.toString)})"
-    case ReadError.AndErrors(list)                => s"AndErrors(${list.map(_.toString)}"
+    case ReadError.MissingValue(path)                   => s"MissingValue(${path})"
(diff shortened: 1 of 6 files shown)


afsalthaj:
Hope we have ran the exact example provided in the issue, https://github.com/zio/zio-config/issues/338

afsalthaj:
Thanks @Dmitrii-Shevchenko !

Issue #356, opened 2020-06-14
Issue in optional fields when comes with missing fields


afsalthaj:
```scala

import java.util.Properties

import zio.config.read

import zio.config.magnolia.DeriveConfigDescriptor._

case class DateVersion(date: String, version: String)

object Main extends App {
  val properties = new Properties()
  properties.setProperty("date", "2016-08-09")
  //properties.setProperty("version", "123")
  
  println(read(descriptor[DateVersion].optional from zio.config.ConfigSource.fromProperties(properties)))
}


```

It returns a None

afsalthaj:
One of the possible solutions is:

```scala

  def loopOptional[B](path: List[Step[K]], keys: List[K], cfg: Optional[B]): Res[Option[B]] =
      loopAny(path, keys, cfg.config) match {
        case Left(error) if hasUnrecoverableErrors(error) =>
          Left(error)
        case Left(AndErrors(_))  => Right(None)
        case Left(error) => Left(error)
        case Right(value) => Right(Some(value))
      }
```

afsalthaj:
There should be a better way than

```

    def loopZip[B, C](path: List[Step[K]], keys: List[K], cfg: Zip[B, C], descriptions: List[String]): Res[(B, C)] =
      (loopAny(path, keys, cfg.left, descriptions), loopAny(path, keys, cfg.right, descriptions)) match {
        case (Right(leftV), Right(rightV)) => Right((leftV, rightV))
        case (Left(leftE), Left(rightE))   => Left(AndErrors(leftE :: rightE :: Nil))
        case (Left(leftE), _)              => Left(ReadError.ForceSeverity(leftE, false))
        case (_, Left(rightE))             => Left(ReadError.ForceSeverity(rightE, false))
      }
```

We need to re-design a few things here

Issue #350, opened 2020-06-05
Improve the ergonomics on defining implicit derivation for config descriptor for custom types


afsalthaj:
As of now, there is a way to define implicit derive instances for custom types. However we should improve the ergonomics and syntactical easiness of this process. More details to follow (typing from mobile device) 

That said, we should also think of having most of the implicit instances(not only for primitives but validations as well) readily available for the user, especially the refined types (which is kept in a separate module). 

More thoughts on refined module in between :
Related to #346 , With inductive implicit resolutions potentially going to be in zio-config-magnolia/shapeless modules , we should have easy integration with refined library. We will keep the verbose module of refined as it is for users who are happy doing the manual config descriptions (or hate magic)

afsalthaj:
Fixed in #358

PR #358, opened 2020-06-17
Bug fixes and improvements

afsalthaj:
- Make listStrict default
- Add more details when error reporting missing values
- Make custom implicit derivations more ergonomic
- Make optional fail if there is atleast one valid value in product
- Make optional fail if the parent path exist and all Childs are missing
- Add documentations for custom derivations in zio-config-magnolia

Fixes the following:
#350 
#356


core/src/main/scala/zio/config/ReadModule.scala
```diff
@@ -81,23 +87,27 @@ private[config] trait ReadModule extends ConfigDescriptorModule {
...
+    def loopZip[B, C](path: List[Step[K]], keys: List[K], cfg: Zip[B, C], descriptions: List[String]): Res[(B, C)] =
+      (loopAny(path, keys, cfg.left, descriptions), loopAny(path, keys, cfg.right, descriptions)) match {
         case (Right(leftV), Right(rightV)) => Right((leftV, rightV))
-        case (Left(leftE), Left(rightE)) =>
-          Left(AndErrors(leftE :: rightE :: Nil))
-        case (Left(leftE), _)  => Left(leftE)
-        case (_, Left(rightE)) => Left(rightE)
+        case (Left(leftE), Left(rightE))   => Left(AndErrors(leftE :: rightE :: Nil))
+        case (Left(leftE), _)              => Left(ReadError.ForceSeverity(leftE, false))
+        case (_, Left(rightE))             => Left(ReadError.ForceSeverity(rightE, false))
```
afsalthaj, line 95: Not a great idea.

Issue #351, opened 2020-06-06
Make listStrict the default behaviour


afsalthaj:
This can be a bit challenging since we need to make sure listStrict don’t break the simple cases of comma separated values in system
Environment or Properties.

senia-psm:
@afsalthaj it might be better to make list behaviour configurable in source, not in description.
It's quite easy to implement.

afsalthaj:
Fixed in #362

PR #362, opened 2020-06-20
Bug fixes, documentation and improvements

afsalthaj:
The PR mainly fixes the following issues:

#350
#356
#351 

1. Make listStrict the default behaviour for type typesafe.config HOCON source, while keeping the behaviour the same as before for other sources.

2. Add more details when error reporting missing values
3. Make custom implicit derivations more ergonomic
4. Make optional fail if there is atleast one valid value in product
5. Add documentations for custom derivations in zio-config-magnolia (more docs)
6. Make the default implicit derivation avoid the name of the name of the SealedTrait

## Following are the breaking changes:

1. List Behaviour

For a HOCON source, `list("key")(string)` used to be a success for non-list value
```scala
 {
    key : value
 }
```

With this PR, this will now on a failure. For HOCON, lists has to be in square brackets.

i.e,

```scala
 {
   key : [value]
 }  
```

All other sources remain the same. i.e, a single value can be considered as list as they work on simple delimiters.

2. Sealed Trait Behaviour

The default derivation of sealed trait don't require the name of the sealed trait to be part of the config

For example:

```scala
sealed trait A 
case object B extends A
case object C extends A
case class D(e: String) extends A
```

```
case class Config(config: A)
```

Preivously `descriptor[A]` requires to have the following HOCON file to successfully parse.
```
{
   config : {
      a : {
        e : "value"
      }
   }
}
```

With this PR, the default behaviour is such that name, the name of the SealedTrait isn't required.

```scala
{
   config : {
        e : "value"
   }
}
```


afsalthaj:
Not sure what exactly is going on with Mdoc
https://app.circleci.com/pipelines/github/zio/zio-config/1117/workflows/d0568dfe-9449-4c5a-be1a-59780499bc5b/jobs/5252/parallel-runs/0/steps/0-103

afsalthaj:
@jdegoes  I am merging these improvements into master, as they pretty much not making any significant behaviour existing apart from the ones mentioned in the PR description.

PR #374, opened 2020-07-08
Remove forced severity errors and revalidate behaviour for optional and default values

### LINKED ISSUE #373: Remove ForcedSeverity errors
The ForcedSeverity was a really helpful hack to avoid errors that happens in option(list) done by  @Dmitrii-Shevchenko. This was aimed to fix the specific issue: https://github.com/zio/zio-config/issues/338

This was a good intermediate solution (a version is released to unblock users who faced this issue). However, it is now time for a more generic solution to default values (such as annotating the behaviour at each node in the tree), that handles other corner cases with default values with a specific spec for options and defaults (and more such behaviours in future)

afsalthaj:
Fixes #373, #356 
Also revisits #356 and apply the fix in a generic way without ForceSeverity error type

Keep a note that tests in "large diff" section in PR:
https://github.com/zio/zio-config/blob/remove_forced_severity/typesafe-magnolia-tests/src/test/scala/zio/config/typesafe/TypesafeConfigOptionalTest.scala


core/src/main/scala/zio/config/ReadError.scala
```diff
@@ -1,28 +1,56 @@
...
-      case ReadError.AndErrors(list)         => list.map(_.size).sum
-      case ReadError.ForceSeverity(error, _) => error.size
+      case e @ ReadError.MissingValue(_, _, _)    => f.applyOrElse(e, (_: ReadError[A]) => alternative)
+      case e @ ReadError.FormatError(_, _, _, _)  => f.applyOrElse(e, (_: ReadError[A]) => alternative)
+      case e @ ReadError.ConversionError(_, _, _) => f.applyOrElse(e, (_: ReadError[A]) => alternative)
+      case e @ ReadError.Irrecoverable(list, _)   => f.applyOrElse(e, (_: ReadError[A]) => go(list))
+      case e @ ReadError.OrErrors(list, _)        => f.applyOrElse(e, (_: ReadError[A]) => go(list))
+      case e @ ReadError.ZipErrors(list, _)       => f.applyOrElse(e, (_: ReadError[A]) => go(list))
+      case e @ ReadError.ListErrors(list, _)      => f.applyOrElse(e, (_: ReadError[A]) => go(list))
+      case e @ ReadError.MapErrors(list, _)       => f.applyOrElse(e, (_: ReadError[A]) => go(list))
```
afsalthaj, line 23: We need to apply the function at each node, before it falls back to alternative or applying fold to each value in the list of errors.

With `alternative` not being present, `case e @ ReadError.Irrecoverable(list, _)  => false` will never pass

core/src/main/scala/zio/config/AnnotatedRead.scala
```diff
@@ -0,0 +1,23 @@
...
+
+  def zip[B](that: AnnotatedRead[B]): AnnotatedRead[(A, B)] =
+    AnnotatedRead((value, that.value), self.annotations ++ that.annotations)
+}
+
+object AnnotatedRead {
+  sealed trait Annotation
+
+  object Annotation {
+    case object NonDefaultValue extends Annotation
```
jdegoes, line 21: I'd change this to `DefaultValue` rather than the negative.

core/src/main/scala/zio/config/ReadError.scala
jdegoes: Ok. In this case, I might pull it out of `ReadError`, since it is not likely to be general-purpose.

core/src/main/scala/zio/config/ReadError.scala
```diff
@@ -132,6 +169,33 @@ sealed trait ReadError[A] extends Exception { self =>
...
+      case ReadError.FormatError(_, _, _, _)  => 1
+      case ReadError.ConversionError(_, _, _) => 1
+      case ReadError.OrErrors(list, _)        => list.map(_.size).sum
+      case ReadError.ZipErrors(list, _)       => list.map(_.size).sum
+      case ReadError.ListErrors(list, _)      => list.map(_.size).sum
+      case ReadError.MapErrors(list, _)       => list.map(_.size).sum
+      case ReadError.Irrecoverable(list, _)   => list.map(_.size).sum
+    }
+
+  def sizeOfZipAndOrErrors: Int =
```
jdegoes, line 185: This one too seems special-purpose.

core/src/main/scala/zio/config/AnnotatedRead.scala
afsalthaj: @jdegoes 

Sure. Just that, we may have some additional logic in ReadModule to handle optional fields.

`NonDefaultValue`  corresponds to field that is either optional or has a default value, but user did provide some value for this field in the source .
Fo example, for the field of type `Option`, `NonDefaultValue` corresponds to `Some(x)` instead of None.

The annotation helps us identify products where user has provided some value for (some of the) optional fields, but forgot to provide the required fields. We fail at the optional node in the program if  `annotations.contains(NonDefaultValue) && numberOfRequiredFields == numberOfMissingValueError`

If we were storing only `DefaultValues`, then we need an additional computation of counting the number of optional fields and we fail if `(numberOfOptionalFields == annotations.filter(_ == NonDefaultValue)` and `numberOfRequiredFields == numberOfMissingValueError`.

afsalthaj: Which I think is fine too

core/src/main/scala/zio/config/ReadError.scala
afsalthaj: Sure, I forgot to move this to Read Module.

afsalthaj: This will also go into the `fold` function

afsalthaj:
Thanks @jdegoes  for reviewing

core/src/main/scala/zio/config/AnnotatedRead.scala
jdegoes: Yes, no worries, and maybe NonDefaultValue is the right choice. Maybe we can just call it `SuppliedValid` or something so it's not the negation of something.

PR #388, opened 2020-07-29
Issue 353 system env delimiters

maciejbak85:
As macros are not doable in a way that it could be used with config.delimiter,
and no need for revolution with for example DelimiterEnumerator,
simple solution was choosen with @afsalthaj 

+ additional change with TypeSafeConfig, will return now ReadError instead of String


maciejbak85:
I see some problem with circleCI config, will fix it

.circleci/config.yml
```diff
@@ -18,6 +18,10 @@ jdk_8: &jdk_8
 jdk_11: &jdk_11
   JDK_VERSION: 11
 
+unit_test_env: &unit_test_env
```
maciejbak85, line 21: needed to unit tests, because code now is using `sys.env ` there.
I think that would be nice to make use of layers there, and instead of getting sys.env directly, take it from ZIO Env.
So will be easier to unit test it.

afsalthaj:
Yes , agree.

typesafe/src/main/scala/zio/config/typesafe/TypesafeConfig.scala
```diff
@@ -3,35 +3,34 @@ package zio.config.typesafe
...
     conf: => com.typesafe.config.Config,
     configDescriptor: ConfigDescriptor[A]
-  )(implicit tag: Tag[A]): Layer[Throwable, Config[A]] =
+  )(implicit tag: Tag[A]): Layer[ReadError[String], Config[A]] =
     Config.fromConfigDescriptorM(
       ZIO
         .fromEither(TypesafeConfigSource.fromTypesafeConfig(conf))
-        .map(configDescriptor from _)
-        .mapError(error => new RuntimeException(error))
+        .bimap(error => ReadError.SourceError(message = error), configDescriptor from _)
```
afsalthaj, line 34: Nice changes here

afsalthaj: I just realised, if  TypesafeConfigSource.fromTypsafeConig could return ReadError

.circleci/config.yml
afsalthaj: 👍

typesafe/src/main/scala/zio/config/typesafe/TypesafeConfigSource.scala
```diff
@@ -22,26 +22,26 @@ object TypesafeConfigSource {
...
+  ): Either[ReadError[String], ConfigSource] =
     Try {
       input
     } match {
-      case Failure(exception) => Left(exception.getMessage)
+      case Failure(exception) => Left(ReadError.SourceError(message = exception.getMessage))
       case Success(value) =>
         getPropertyTree(value) match {
-          case Left(value)  => Left(value)
+          case Left(value)  => Left(ReadError.SourceError(message = value))
```
afsalthaj, line 44: Perfect!

core/src/main/scala/zio/config/ReadError.scala
```diff
@@ -106,9 +108,17 @@ sealed trait ReadError[A] extends Exception with NoStackTrace { self =>
         )
       )
 
+    def renderSourceError[A](err: ReadError.SourceError[A]): Sequential =
+      Sequential(
+        List(
+          Failure(s"SourceError: ${err.message}" :: Nil)
+        )
+      )
```
afsalthaj, line 116: Yea, this makes sense too. Coz source can be per field as well and can be part of accumulated read errors!

core/src/main/scala/zio/config/ConfigSourceModule.scala
```diff
@@ -380,12 +378,17 @@ trait ConfigSourceStringModule extends ConfigSourceModule {
...
       valueDelimiter: Option[Char],
       leafForSequence: LeafForSequence = LeafForSequence.Valid
-    ): UIO[ConfigSource] =
-      UIO
-        .effectTotal(sys.env)
-        .map(
-          map => fromMap(map, SystemEnvironment, keyDelimiter, valueDelimiter, leafForSequence)
-        )
+    ): IO[ReadError[String], ConfigSource] = {
+      val validDelimiters = ('a' to 'z') ++ ('A' to 'Z') :+ '_'
```
afsalthaj, line 382: Its a tight constraint.

afsalthaj:
Thanks @maciejbak85 for this change.

Issue #389, opened 2020-07-29
Remove call to sys.env and use ZLayer


maciejbak85:
According to discussion in PR: 
https://github.com/zio/zio-config/pull/388

would be good to make a change in ConfigSourceModule fromSystemEnv.
Change calling to external state `IO.effectTotal(sys.env)` and take env from a Layer.

This change will make this code more functional ( as no relying on external state ), and easier to unit test.
Currently in circleCI config there had to be set system variables, to make unit tests working.
So after using layer there, they can be removed from there.

maciejbak85:
I would like to work on it ofc :)

jdegoes:
@maciejbak85 I approve! Would be great to require that in the layer. Also for anything else where ZIO Config is touching the real world, would be better to grab that in the layer so it can be better tested.

PR #390, opened 2020-07-30
Issue 389 sys env layer

maciejbak85:
The same can be done with properties.
Making it this way we gain:

1. code more functional, easy to test
2. library user still can use predefined code for getting system envs ( by SystemModule )

we can do the same with Properties.
what do you think about this change ?

https://github.com/zio/zio-config/issues/389


maciejbak85:
@afsalthaj what do you think about this change ?

afsalthaj:
@maciejbak85 thanks  for working on this. Give me some time thinking about it.

core/src/main/scala/zio/config/ConfigSourceModule.scala
```diff
@@ -378,12 +379,16 @@ trait ConfigSourceStringModule extends ConfigSourceModule {
       keyDelimiter: Option[Char],
       valueDelimiter: Option[Char],
       leafForSequence: LeafForSequence = LeafForSequence.Valid
-    ): IO[ReadError[String], ConfigSource] = {
+    ): ZIO[SystemModule, ReadError[String], ConfigSource] = {
```
afsalthaj, line 382: I think its a good idea, and gels well with ZIO's Environment.
 
Should we need another function along with this that return IO?, because I find a few users who like just using ZIO's IO bifunctor, probably by providing the system layer already, allowing them to do

```
for {
  env <- ConfigSource.fromSystemEnv
  prop <- ConfigSource.fromSystemProperties
  config. <- read(descriptor[AppConfig] from env.orElse(prop))
} yield config

```

core/src/main/scala/zio/config/ConfigStringModule.scala
```diff
@@ -382,7 +384,7 @@ trait ConfigStringModule extends ConfigModule with ConfigSourceStringModule {
       configDescriptor: ConfigDescriptor[A],
       keyDelimiter: Option[Char] = None,
       valueDelimiter: Option[Char] = None
-    )(implicit tag: Tag[A]): Layer[ReadError[String], Config[A]] =
+    )(implicit tag: Tag[A]): ZLayer[SystemModule, ReadError[String], Config[A]] =
```
afsalthaj, line 387: This is minor, but worth noting. All other functions in this module returns a Layer.
That said, I realise myself that not all functions in `ConfigSourceModule` returned `IO` before.

core/src/main/scala/zio/config/SystemModule.scala
```diff
@@ -0,0 +1,36 @@
...
+        val prop = new Properties
+        propMap.map {
+          case (key, value) => prop.setProperty(key, value)
+        }
+        prop
+      }
+    })
+
+  def getEnvironment: ZIO[SystemModule, Throwable, Map[String, String]] =
+    ZIO.accessM(_.get.getEnvironment)
```
afsalthaj, line 34: Is this currently used only in tests?

afsalthaj:
I think it’s a cool change. Whenever you are ready, and one minor change suggested in ConfigSourceModule, we are good to merge.

core/src/main/scala/zio/config/SystemModule.scala
maciejbak85: yes

core/src/main/scala/zio/config/ConfigStringModule.scala
maciejbak85: hmm you want me to change return types for others to ZLayer[Any then ? :)

core/src/main/scala/zio/config/ConfigSourceModule.scala
maciejbak85: indeed - should be usefull, how should we name it then ?
Maybe: fromSystemEnvUnsafe ?
I need to think about proper name here then

maciejbak85:
ah I just realized that ZIO has already System module.. ;) will use it instead of new module that Ive added here

core/src/main/scala/zio/config/ConfigStringModule.scala
maciejbak85: btw not all:

 ```
def fromSystemProperties[K, V, A](
      configDescriptor: ConfigDescriptor[A],
      keyDelimiter: Option[Char] = None,
      valueDelimiter: Option[Char] = None
    )(implicit tag: Tag[A]): ZLayer[System, ReadError[String], ZConfig[A]]
```
is returning ZLayer here already

core/src/main/scala/zio/config/ConfigSourceModule.scala
maciejbak85: Ok Ive choosen: fromSystemEnvLive, hope its fine

afsalthaj:
I will review this today

afsalthaj:
@maciejbak85 whenever we are ready we are good to merge this one

maciejbak85:
> @maciejbak85 whenever we are ready we are good to merge this one

can we merge it ? last days I was quite busy with my work :) I am still, but not sure if I need to finish here smth more still ( definitely update branch )

lemastero (approved):

lemastero:
@afsalthaj because @maciejbak85 is a bit busy, I helped with a broken CI.
Now PR is ready to be merged, whenever you are ready.

afsalthaj:
@lemastero Thanks for taking this up.

afsalthaj (approved):

PR #396, opened 2020-08-04
Handle duplicate keys in the source when convert to Hocon

### LINKED ISSUE #393: Handle duplicate keys in the source when writing back to Json
Make sure write back to Json handle duplicate keys.

In the below example, you can see key `metadata` happens twice.

```
  final case class A(metadata: String, c: String)

  final case class B(metadata: List[A])

  val success =
    B(List(A("abc", "xyz")))

  println(write(descriptor[B], success).map(t => t.toJson))

```

returns

```scala
Right({
    "metadata" : [
        "abc"
    ]
}
)

```

It misses out information about c.

A fix from the user's side is renaming `metadata` to something else

```
  final case class A(metadata: String, c: String)

  final case class B(list: List[A])

  val success =
    B(List(A("abc", "xyz")))

  println(write(descriptor[B], success).map(t => t.toJson))

```

will return the actual result

```scala
Right({
    "list" : [
        {
            "c" : "xyz",
            "metadata" : "abc"
        }
    ]
}
)

```

tuleism:
Fix #393


afsalthaj:
@tuleism Thanks for your help in fixing the issue. Nice one 👍

PR #441, opened 2020-11-15
Fix generateDocs for some recursive cases

vigoo:
Fix recursion support in `generateDocs` when there is a dynamic `XmapEither` directly under `Nested` that prevented detection of the loop. Added a test case that demonstrates this.


core/src/main/scala/zio/config/ConfigDocsModule.scala
```diff
@@ -420,6 +420,12 @@ trait ConfigDocsModule extends WriteModule {
    * from various sources.
    */
   final def generateDocs[A](config: ConfigDescriptor[A]): ConfigDocs = {
+    def lookAhead(desc: LazyConfigDescriptor[_]): ConfigDescriptor[_] =
```
afsalthaj, line 423: I see where this is going. Nice one on understanding this verbose logic and fixing the issue. Thanks @vigoo 

FYI (and probably unrelated): 

I am making some very minor changes such that `LazyConfigDescriptor ` becomes one of the terms in the ADT, but rest of the code that makes recursion work remains the same. 

https://github.com/zio/zio-config/blob/api_docs/core/src/main/scala/zio/config/ConfigDescriptorModule.scala#L1317

core/src/test/scala/zio/config/RecursiveConfigTest.scala
```diff
@@ -110,4 +169,31 @@ object RecursiveConfigTestUtils {
...
+            case _          => None
+          }
+        )
+    )
+
+    lit <> add
+  }
+
+  val exprValue  = Add(List(Lit(1), Add(List(Add(List(Lit(2), Lit(3))), Lit(4))), Lit(5)))
+  val exprSource = ConfigSource.fromPropertyTree(write(expr, exprValue).toOption.get, "test", LeafForSequence.Invalid)
```
afsalthaj, line 198: @vigoo  Will this fix the issue in your zio-aws docs?

afsalthaj (approved):

core/src/test/scala/zio/config/RecursiveConfigTest.scala
vigoo: I don't think so, will look into that when I have time.

afsalthaj:
If you could share the config here I can take a look at it as well

vigoo:
https://github.com/vigoo/zio-aws/blob/master/zio-aws-core/src/main/scala/io/github/vigoo/zioaws/core/config/package.scala#L174-L540 

this is it, sorry I could not get back to work on it yet

afsalthaj:
@vigoo  Currently I am looking at zio-aws configuration. It looks good the way it is written.

I will have a few notes as part of this issu. For example: zio-config cannot know the constant values such as `default`, `anonymous` etc when it comes to fetching credentials provider. A more detailed explanation (and alternatives) will be given as soon as possible.

afsalthaj:
And may be we will be fine with the current implementation as far as I can see. I will give a confirmation at a later point though.
Hope you don't mind me raising a PR to zio-aws. Happy to contribute.

vigoo:
Thanks for checking it and raising issues/PRs! I will try to get back to this topic soon just a bit busy these days.

afsalthaj:
No worries. The current implementation of markdown is fairly complex
because of look-ahead logic but works for the most part. I hope any change,
if needed, will be a minor change. Most of them can be solved if config is
enriched with more static types instead of embedding them into xmapEither
logic (where zio-config cannot pre-document these info that are available
only at runtime).

I haven’t played much with recursion config doc, but looks good as far as I
can tell from your tests. Really useful bit!

PS: Today John merged a big PR of mine (62 files),  jus make sure you
merge. There can be some conflicts.

vigoo:
The fix does not work with Scala 2.11 and I don't have more time now to figure that out. I believe it is because of different equality behavior in 2.11 for lambdas. This whole reentrance detection is quite fragile this way but I have no better idea right now. 
The tests pass for 2.12 and 2.13

afsalthaj:
> The fix does not work with Scala 2.11 and I don't have more time now to figure that out. I believe it is because of different equality behavior in 2.11 for lambdas. This whole reentrance detection is quite fragile this way but I have no better idea right now.
> The tests pass for 2.12 and 2.13

I have rerun the 211 build and it worked. Or do you mean it fails at runtime for scala 211 for any of your examples?

afsalthaj:
Thanks Daniel for the PR!

vigoo:
I disabled the failing test for 2.11 with a test aspect so you can decide
to have this partial fix at least or not.

afsalthaj (approved):

PR #451, opened 2020-11-27
Simplify manual config-desciption of sealed-traits (coproducts)

### LINKED ISSUE #450: Simplify manual config-desciption of sealed-traits (coproducts) without relying on zio-config-magnolia
As of now, users find it really hard to describe `ConfigDescriptor` when the config consist of `Coproducts` (sealed-trait) in it, unless they are depending on auto-derivation using `zio-config-magnolia`. 

This is because of the complexity that is involved in reading the config and writing them back individually, such that each term during its write phase can fail if the instance isn't matching. Later on users need to make use `orElse` combinator to finally form the coproduct type.

This logic is automatically handled in `zio-config-magnolia` making auto-derivation as a go-to approach for configs involving coproducts. However, we need a solution that allows us to describe coproducts (i.e, summon ConfigDescriptor) manually without having to rely on zio-config-magnolia

afsalthaj:
Fix #450


core/src/main/scala/zio/config/ConfigDescriptorModule.scala
```diff
@@ -1206,6 +1207,177 @@ trait ConfigDescriptorModule extends ConfigSourceModule { module =>
...
+     *
+     * {{{
+     *   import zio.config.magnolia._
+     *
+     *   val config = descriptor[D]
+     * }}}
+     */
+    def enum[D] = new PartiallyAppliedEnum[D]
+
+    class PartiallyAppliedEnum[D] {
```
afsalthaj, line 1249: To avoid users having to give types for all subtypes

core/src/main/scala/zio/config/ConfigDescriptorModule.scala
```diff
@@ -1206,6 +1207,177 @@ trait ConfigDescriptorModule extends ConfigSourceModule { module =>
...
+     *
+     * Example:
+     *
+     * {{{
+     *   import zio.config.magnolia._
+     *
+     *   val config = descriptor[D]
+     * }}}
+     */
+    def enum[D] = new PartiallyAppliedEnum[D]
```
jdegoes, line 1247: We should call this `enumeration` rather than `enum`, because in Scala 3, `enum` is a keyword.

jdegoes (changes requested):
Good to merge after `enum` is renamed to `enumeration`. Looks great!

afsalthaj:
@jdegoes Thanks John.
I have renamed it. I don't have merge permission without an approval

jdegoes (approved):

PR #452, opened 2020-11-28
Add zio-config-gen for easier generation of an (almost) valid config for user to edit

### LINKED ISSUE #453: Create a gen module to generate Config for user to edit instead of having to write from scratch using documentation

afsalthaj:
Fixes #453 

```scala

 import zio.config._, magnolia._

 sealed trait Region

 @name("ap-southeast")
 case object ApSouthEast2 extends Region

 @name("usEast")
 case object UsEast extends Region

 final case class Database(port: Int, host: java.net.URL)
 final case class MyConfig(region: Region, database: Database)

 val result =
   generateConfigJson(descriptor[MyConfig], 2).unsafeRunChunk

 println(result)

```

yields

```scala

Chunk({
    "database" : {
        "host" : "http://def",
        "port" : "7300"
    },
    "region" : "ap-southeast"
}
,{
    "database" : {
        "host" : "http://abc",
        "port" : "8908"
    },
    "region" : "usEast"
}
)


```


core/src/main/scala/zio/config/ConfigStringModule.scala
```diff
@@ -244,7 +244,7 @@ trait ConfigStringModule extends ConfigModule with ConfigSourceStringModule {
     def file(path: String): ConfigDescriptor[File] = nested(path)(file)
 
     val float: ConfigDescriptor[Float] =
-      source(ConfigSource.empty, PropertyType.FloatType) ?? "value of type float"
+      sourceDesc(ConfigSource.empty, PropertyType.FloatType) ?? "value of type float"
```
jdegoes, line 247: Why all the name changes, adding `Desc` as suffix?

afsalthaj:
They are not exposed to users, but constructors that ensures lazyDesc is
being used internally. Example nestedDesc ensures Nested is being called
with lazyDesc.

For these constructors, I thought of not using the same name as those
functions exposed to the user. Example: “nested” that is available to the
user through zio.config.ConfigDescriptor._

Another reason is, I couldn’t also use “lazy” because it’s a key word and I
had to use lazyDesc. So to keep the consistency I thought of sufficing Desc
to all constructor functions.

core/src/main/scala/zio/config/ConfigStringModule.scala
afsalthaj: They are not exposed to users, but constructors that ensures lazyDesc is being used internally. Example nestedDesc ensures Nested is being called with lazyDesc. 

For these constructors, I thought of not using the same name as those functions exposed to the user. Example: “nested” that is available to the user through zio.config.ConfigDescriptor._

Another reason is, I couldn’t also use “lazy” because it’s a key word and I had to use lazyDesc. So to keep the consistency I thought of sufficing Desc to all constructor functions.

vigoo:
I checked this as part of our discussion in https://github.com/vigoo/zio-aws/issues/144 and wanted to question a bit the design, sorry if I don't see the whole picture or vision here ;) 

So if I'm correct the idea is the following:
- There is a `DeriveGen` type class wrapping a `zio.test.Gen` with auto derivation
- With `DeriveGen[A]` and `ConfigDescriptor[A]` we can generate an arbitrary `A` then write it to a `PropertyTree` and serialize to any of the supported config formats

This is already very useful as is, but I think it's a bit limited from user pespective and also will be harder to maintain. I think the generator instance should be part of the config description. That way:
- `ConfigDescriptor` will remain the single source of truth and a full description of your configuration
- It matches the nice composable API of config descriptors. You would be able to attach and manipulate generators on any level
- The magnolia derivation can use the same operators to attach the generators on the fly when it derives the config descriptor. This would increase compilation time (one derivation instead of two) and I think would also make future developments easier, as in the current case if some derivation-changing annotations would be added it possibly has to be implemented in both places. (as an example think of something like `@ignore` to ignore a constructor of an ADT) 
- You get more expression power when writing custom descriptors as you can have different generators for the same type per use case
- All the default config descriptor builder functions could set up the default generators defined in the PR without any macro or even implicit resolution needed.

To sum it up I believe this would be cleaner and more generic. What do you think?

afsalthaj:
@vigoo I think, the idea of grabbing the information about generators as part of ConfigDescriptor increases the flexibility to manipulate the generation.  It is definitely an improvement. 

As of now, DeriveGen randomly generates some value, but it does go through the “write” back functionality of the given ConfigDescriptor nodes producing a decent valid config for the user to start with, without any additional work. When we add composable generator nodes in ConfigDescriptor, we can  use them as an additional piece of info along with or overriding DeriveGen instances. 

However we should give user an interface to generate without defining zero generator nodes in their ConfigDescriptor program.  So I believe it should be a mix of DeriveGen and
generator nodes in ConfigDescriptor to allow flexibility. 

I am not fully sure of the maintainability issue though. I imagined it as something written once off. We hardly want to change anything here the way it is now.

jdegoes (approved):

Issue #471, opened 2020-12-11
Add a filter function on ConfigSource


fsarradin:
If my key looks like this `key1.key2.key3`, and my env var looks like `P-KEY1_KEY2_KEY3`, `.convertKeys` on `ConfigSource` doesn't help here. Typically, `.convertKeys(k => addPrefixToKey("P-")(k.toUpperCase))` will only match `P-KEY1_P-KEY2_P-KEY3`.

In addition, if I want `P_` (instead of `P-`) as a prefix, it won't work, as `_` is used as a prefix delimiter.

The idea would be to add a `filter` function on `ConfigSource`.

The current workaround is to load env vars in a Map and transform them before using them in zio-config through `zio.config.fromMap`.

PR #475, opened 2020-12-23
WIP: PropertyTree Optimizations

### LINKED ISSUE #418: Very slow performance with large number of env vars.
When you run your app on Kubernetes there are a lot of env vars defined. At least two for each service. So in my case there are 800 of env vars like `LONG_SERVICE_NAME_HOST` and `ANOTHER_LONG_SERVICE_NAME_PORT`. 

And it seems that config loading is stuck forever in `PropertyTree#mergeAll`.

tusharmath:
Should Fix #418

### DIFF
--- a/core/src/main/scala/zio/config/PropertyTree.scala
+++ b/core/src/main/scala/zio/config/PropertyTree.scala
@@ -115,14 +115,17 @@ sealed trait PropertyTree[+K, +V] { self =>
    */
   final def merge[K1 >: K, V1 >: V](that: PropertyTree[K1, V1]): List[PropertyTree[K1, V1]] =
     (self, that) match {
-      case (Sequence(l), Sequence(r)) => singleton(Sequence(l ++ r))
+      // DEOPT: O(n) time/space complexity operation in "l ++ r"
+      case (Sequence(l), Sequence(r))           => singleton(Sequence(l ++ r))
       case (l: Record[K, V], r: Record[K1, V1]) =>
+        // DEOPT: O(n) time/space complexity operation in "l.value.keySet ++ r.value.keySet"
         (l.value.keySet ++ r.value.keySet)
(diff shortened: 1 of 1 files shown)


afsalthaj:
This PR is promising @tusharmath.
Thanks for introducing thoughts on asymptotic analysis into PropertyTree. 
Looking forward

tusharmath:
**Some initial thoughts**

1. Consider a property tree of 10 keys being merged with another of 10 keys. The total number of options produced will be equal to `2^10 = 1024`. 
2. Clearly the memory requirement also increases exponentially and using a mutable data structure might not really help.
3. I think a more lazy solution might actually perform better.
4. Alternatively we can create a new API to merge two tree with a conflict resolver, such that instead of returning a list it returns a single merge property tree.

_Please correct me if I am wrong_

afsalthaj:
Merging trees should produce multiple possibilities returning a list. Changing that can cause issues in future.

A laziness approach would be rather better but we need to see how that goes. I was pretty busy with a few other changes, but happy to work with you on this.

Meanwhile, as a first step (or as a minimum requirement) we need a filter function that allows us to fetch a trimmed down version of the tree especially for sources such as system-env, properties etc. `convertKeys` & `mapKeys` are playing on an already-fetched tree, and hence this change need to be within each source. This is related to https://github.com/zio/zio-config/issues/471

afsalthaj:
Also, probably merging trees should produce only distinct trees. I think we should ensure that as well as the next stage improvement

afsalthaj:
Hi @tusharmath 
How are we going about this issue ? (While I know you have been super busy with some wonderful contributions in zio space)

tusharmath:
Haven't been able to spend much time on this. Apologies for that. Please
feel free to assign it to yourself.

afsalthaj:
@tusharmath No worries.

PR #476, opened 2020-12-24
Solve recursive config structure by ensuring object identity & building program-summary to keep the deep error reports.

afsalthaj:
The change is to make zio-config read, write and document recursive config structures. 
The main idea of ensuring object identity seem to be working with the introduction of more laziness across.

 - [x] All tests related to recursion is passing now.
 
image

 - [x] All tests remaining in zio-config that has strong semantics on error reporting is passing now. The main requirement was to keep all deep error reports especially on optional configs
 
image

 - [x] Add tests for auto derivation of recursive configs (especially typesafe)
 - [x] Fix the stack overflow error when computing the sources in a config-descriptor.
 - [x] Distinguish between re-use of products/coproducts and recursion. Otherwise, users miss out deep error reports if there is a repetition of a structure in the config.
 - [x] A bit of concern on performance, but I don't have any numbers yet. Some tests such as `collectAllRoundTripTest` did fail with timeouts during the initial commits in this PR, however these are fixed later on. The main reason was due to by-name configsource.


core/src/main/scala/zio/config/ConfigDescriptorModule.scala
```diff
@@ -998,6 +998,41 @@ trait ConfigDescriptorModule extends ConfigSourceModule { module =>
       loop(self)
     }
 
+    lazy val sources: Set[ConfigSource] = {
```
afsalthaj, line 1001: Sources is yet to be used.

afsalthaj: Most of the logic is just to check if object identity works for recursive structure

project/BuildHelper.scala
```diff
@@ -19,7 +19,7 @@ object BuildHelper {
   )
 
   private val std2xOptions = Seq(
-    "-Xfatal-warnings",
+    //"-Xfatal-warnings",
```
afsalthaj, line 22: Only because we wanted more console-level testing

core/src/test/scala/zio/config/RecursiveConfigTest.scala
```diff
@@ -1,28 +1,30 @@
...
+import zio.test.TestAspect.ignore
 
-import zio.config.ConfigDescriptor._
+import zio.config.ConfigDescriptor._, ConfigDescriptorAdt._
 
 object RecursiveConfigTest
     extends BaseSpec(
       suite("RecursiveConfigTest")(
         test("read simple") {
+          // FIXME: The logic works, more work required in cleaning up the logic.
```
afsalthaj, line 16: With this PR, we are making this tests work

core/src/main/scala/zio/config/ReadModule.scala
```diff
@@ -232,24 +244,33 @@ private[config] trait ReadModule extends ConfigDescriptorModule {
       path: List[Step[K]],
       keys: List[K],
       config: ConfigDescriptor[B],
-      descriptions: List[String]
-    ): Res[B] =
+      descriptions: List[String],
+      accompaniedComputationFailed: Boolean
+    ): Res[B] = {
+      programSummary = config :: programSummary
```
afsalthaj, line 250: There is a logical error here. Not all config needs to be added

core/src/main/scala/zio/config/ReadModule.scala
```diff
@@ -232,24 +244,33 @@ private[config] trait ReadModule extends ConfigDescriptorModule {
...
-        case c @ OrElseEither(_, _)       => loopOrElseEither(path, keys, c, descriptions)
+        case Lazy(thunk)       => loopAny(path, keys, thunk(), descriptions, accompaniedComputationFailed)
+        case c @ Default(_, _) => loopDefault(path, keys, c, descriptions, accompaniedComputationFailed)
+        case c @ Describe(_, message) =>
+          loopAny(path, keys, c.config, descriptions :+ message, accompaniedComputationFailed)
+        case c @ DynamicMap(_, _) => loopMap(path, keys, c, descriptions, accompaniedComputationFailed)
+        case c @ Nested(_, _, _)  => loopNested(path, keys, c, descriptions, accompaniedComputationFailed)
+        case c @ Optional(_)      =>
+          // FIXME: Temporary logic to check if identity can help us short circuit recursive trees
+          if (programSummary.exists(_.equals(c)) && accompaniedComputationFailed)
```
afsalthaj, line 260: Shouldn't be in Optional. Should be outside. This is wrong

core/src/main/scala/zio/config/ConfigDescriptorModule.scala
```diff
@@ -1880,49 +1915,77 @@ trait ConfigDescriptorModule extends ConfigSourceModule { module =>
...
-      Nested(source, path, lazyDesc(config))
+    final def nestedDesc[A](source: ConfigSource, path: K, config: => ConfigDescriptor[A]): ConfigDescriptor[A] = {
+      lazy val config0 = config
+      Nested(source, path, lazyDesc(config0))
+    }
 
-    final def optionalDesc[A](config: => ConfigDescriptor[A]): ConfigDescriptor[Option[A]] =
-      Optional(lazyDesc(config))
+    final def optionalDesc[A](config: => ConfigDescriptor[A]): ConfigDescriptor[Option[A]] = {
+      lazy val config0 = config
```
afsalthaj, line 1947: Probably not required

core/src/main/scala/zio/config/ConfigDescriptorModule.scala
```diff
@@ -976,28 +972,133 @@ trait ConfigDescriptorModule extends ConfigSourceModule { module =>
...
         config match {
-          case Lazy(thunk)                  => Lazy(() => loop(thunk()))
-          case Source(source, propertyType) => Source(f(source), propertyType)
-          case DynamicMap(source, conf)     => DynamicMap(f(source), loop(conf))
-          case Nested(source, path, conf) =>
+          case c @ Lazy(thunk) =>
+            val res = thunk()
+            summary.keys.find(_ == c) match {
+              case Some(value) =>
+                summary(c).asInstanceOf[ConfigDescriptor[B]]
```
afsalthaj, line 989: Ha! Thats `.get(c) match`

core/src/main/scala/zio/config/AnnotatedRead.scala
```diff
@@ -18,5 +18,6 @@ object AnnotatedRead {
   sealed trait Annotation
   object Annotation {
     case object NonDefaultValue extends Annotation
+    case object Recursive       extends Annotation
```
afsalthaj, line 21: Not required anymore

afsalthaj: With making `lazy val sources` working now, remove `Recursive` annotations. This was part of a hacky solution as part of testing

core/src/main/scala/zio/config/ReadModule.scala
```diff
@@ -234,22 +230,75 @@ private[config] trait ReadModule extends ConfigDescriptorModule {
...
     loopAny(Nil, Nil, configuration, Nil).map(_.value)
+
+  }
+
+  private[config] def isEmptyConfigSource[A](
+    config: ConfigDescriptor[A],
+    keys: List[K]
+  ): Boolean = {
+    val sourceTrees = config.sources.map(_.getConfigValue(keys))
+    sourceTrees.forall(_ == PropertyTree.empty)
```
afsalthaj, line 301: PropertyTree.Empty is the only condition we need to verify and not `PropertyTree.isEmpty` because, `PropertyTree.isEmpty` returns true for empty `Sequence`

afsalthaj:
@jdegoes Could you take a first level review ?
A few tests are failing. The explanations for them are given in the PR description. In short, they are not major roadblock and I will be fixing them too, going forward.

afsalthaj:
With the latest commit (87164eb), all logical errors are fixed.
The only ones that are failing is due to slow tests and has been confirmed to working for smaller inputs. This is related to performance hiccups after the introduction of more computation in read module. We will need to fix this performance issue

core/src/main/scala/zio/config/ReadModule.scala
```diff
@@ -232,24 +239,77 @@ private[config] trait ReadModule extends ConfigDescriptorModule {
...
-        case c @ Optional(_)              => loopOptional(path, keys, c, descriptions)
-        case c @ OrElse(_, _)             => loopOrElse(path, keys, c, descriptions)
-        case c @ OrElseEither(_, _)       => loopOrElseEither(path, keys, c, descriptions)
-        case c @ Source(_, _)             => loopSource(path, keys, c, descriptions)
-        case c @ Zip(_, _)                => loopZip(path, keys, c, descriptions)
-        case c @ TransformOrFail(_, _, _) => loopXmapEither(path, keys, c, descriptions)
-        case c @ Sequence(_, _)           => loopSequence(path, keys, c, descriptions)
+      // FIXME: Distinguish between re-use of structures vs recursive structures.
+      // If not distinguished, we miss out deep errors in the error report when config has same products/co-products repeated (than a recursion).
+      // However, this is not a major roadblock, as we keep deep error-reports in all other cases.
```
afsalthaj, line 247: This is done. Remove the comments

core/src/main/scala/zio/config/ReadModule.scala
```diff
@@ -232,24 +239,77 @@ private[config] trait ReadModule extends ConfigDescriptorModule {
...
+      // {{{
+      //   MissingValue
+      //     path: a.a.b2
+      //     Details: optional value, value of type string
+      //   MissingValue
+      //     path: a.a.a2
+      //     Details: optional value, value of type string
+      // }}}
+      //
+      if (isEmptyConfigSource(config, keys.reverse) && programSummary.contains(config)) {
```
afsalthaj, line 273: To avoid calculating `source`, use `programSummary.contains(config)` first

core/src/test/scala/zio/config/RecursiveConfigTest.scala
```diff
@@ -113,34 +137,166 @@ object RecursiveConfigTest
               )
             )
           )
-        } @@ TestAspect.exceptScala211
+        } @@ ignore @@ TestAspect.exceptScala211
```
afsalthaj, line 140: Remove ignore

core/src/main/scala/zio/config/ReadModule.scala
```diff
@@ -232,24 +239,61 @@ private[config] trait ReadModule extends ConfigDescriptorModule {
       path: List[Step[K]],
       keys: List[K],
       config: ConfigDescriptor[B],
-      descriptions: List[String]
+      descriptions: List[String],
+      programSummary: List[ConfigDescriptor[_]]
```
afsalthaj, line 243: An early exit should be only for recursive values (`programSummary.contains(config)`)
  Such that if configuration isn't recursive, we report all errors at deeper levels.
  
  Example:
  Without a program summary, for a non-recursive structure, we get the following error structure:
 ```scala
     MissingValue
      path: a.a
      Details: optional value
   ```
  
  With program summary, for a non-recursive structure, we get the following error structure
  
  ```scala
    MissingValue
      path: a.a.b2
      Details: optional value, value of type string
    MissingValue
      path: a.a.a2
      Details: optional value, value of type string
  ```

jdegoes (approved):

PR #478, opened 2021-01-01
Cross-compile to Javascript using Scala-JS and Shocon

nightscape:


nightscape:
@afsalthaj the big question I have regarding this PR is if one wants to restrict zio-config to the subset of functionality that is available on both the JVM and JS, or if both platforms can have platform-specific functionality.
In Shocon you e.g. cannot specify an arbitrary config file, instead it parses an `application.conf` file at compile time and provides the pre-parsed values at run-time.
Other things that cause problems are `java.net.URL`s (`java.net.URI` works though), `java.io.File`s and `java.nio.file.Path`.

nightscape:
I'm also running into many of the problems mentioned here:
https://github.com/akka-js/shocon/issues/10

afsalthaj:
@nightscape Yes it's a concern and I can see Shocon solved it through macros. I will need some time to think whether we should have such a capability. 
Thanks for your PR. This is great 👍

nightscape:
I've started extracting `trait`s with the behavior common to both JS and JVM and then having concrete `class`es/`object`s extending them with platform specific behavior. I don't know if this works in all cases, but in the cases where it does, one does not have to relinquish existing functionality.

KadekM:
This would be nice to have :-)

nightscape:
I assume this has been superseded by https://github.com/zio/zio-config/pull/530, right @afsalthaj ?

PR #500, opened 2021-03-14
Support for Scala3 and simplifying ConfigDescriptor with the introduction of `to` instead of `apply` and `unapply`

### LINKED ISSUE #182: Cross-compile for Dotty

vigoo:
Resolves #182 

Currently crashes the Scala 3 compiler, to be investigated.


vigoo:
Dotty crash minimized in https://github.com/lampepfl/dotty/issues/11732

vigoo:
Note: infinite recursion in `zio.config.ConfigDocsModule.loop` with Scala 3

core/src/test/scala/zio/config/NestedConfigTest.scala
```diff
@@ -67,15 +68,16 @@ object NestedConfigTestUtils {
...
       val database =
-        (string("dburl")(DbUrl.apply, DbUrl.unapply)
+        (string("dburl")
+          .to[DbUrl]
           .orElseEither(nested("connection")(dbConnection)) |@|
-          nested("credentials")(credentials).optional)(Database.apply, Database.unapply)
+          nested("credentials")(credentials).optional).to[Database]
 
-      (nested("database")(database) |@| double("pricing"))(AppConfig, AppConfig.unapply)
+      (nested("database")(database) |@| double("pricing")).to[AppConfig]
```
afsalthaj, line 80: No a PR comment since it is in very early state, but curious:
I am just searching for `.to[CaseObject]`. Is there one anywhere?

vigoo: Hm I think that would be simple to support in the scala2 macro but I have no idea how to do it in scala3, but worths a try

vigoo:
The infinite recursion with Scala3 happens with the test that does this:

```scala
def expr: ConfigDescriptor[Expr] = {
    val lit: ConfigDescriptor[Expr] = int.transformOrFail(
      n => Right(Lit(n)), {
        case Lit(n) => Right(n)
        case _      => Left(s"Not Lit")
      }
    )

    val add: ConfigDescriptor[Expr] = nested("add")(
      list(expr)
        .apply(
          lst => Add(lst), {
            case Add(items) => Some(items)
            case _          => None
          }
        )
    )

    lit <> add
  }

generateDocs(expr).toTable
```

To be honest when I added recursion the first time I felt this termination check is a bit fragile and still feel :) For example I don't even understand how this can work with Scala 2.x, as `expr` is a `def`. Hard to tell what exactly "stops" the recursion.

So the first thing I would do is to not support this but change it to a `lazy val expr = ...`. 
That way we should be able to say that `expr` has a fix value and we can detect the reentry to it when traversing the config descriptors.

This still does not fix it in Scala 3 though, because the current implementation does not check reentrance on each step only in case of `Nested`, with some lookahead. In this example this means that with Scala 2.x, the value produced by `list(expr)` in `add` is somehow stable and detected as recursion, while in Scala 3 it is a different `Sequence` node each time.

So I changed this whole thing a bit and now checking each descriptor if it has been already "seen" and stop if it has.

This of course change a bit the output too, but I'm not sure if the previous one was correct or the best possible.

Previously with Scala 2.x this test case produced the following docs (in markdown):

|FieldName|Format                         |Description|Sources|
|---      |---                            |---        |---    |
|         |[any-one-of](fielddescriptions)|           |       |

### Field Descriptions

|FieldName |Format     |Description      |Sources|
|---       |---        |---              |---    |
|          |primitive  |value of type int|       |
|[add](add)|[list](add)|                 |       |

### add

|FieldName|Format   |Description      |Sources|
|---      |---      |---              |---    |
|         |primitive|value of type int|       |
|add      |recursion|                 |       |

Which is true, but a bit redundant in my opinion.

With my changes it produces the following one with both Scala 2.x and Scala 3:

|FieldName|Format                         |Description|Sources|
|---      |---                            |---        |---    |
|         |[any-one-of](fielddescriptions)|           |       |

### Field Descriptions

|FieldName|Format           |Description      |Sources|
|---      |---              |---              |---    |
|         |primitive        |value of type int|       |
|add      |list of recursion|                 |       |

Now this is more concise but probably less understandable what the recursion means.

I don't "feel" that I much the requirements of the documentation generator feature so @afsalthaj you could help figuring out / fixing this to work like you would want to.

afsalthaj:
@vigoo  Read your previous commit.
During the time I revisited recursion in ReadModule last year (#476 , I hardly got my hands on to ConfigDocs module to handle anything related to Recursion, as the specs worked straight away. I remember I was having this mixed feeling of  "Well  that reduced a bit of my work that somehow recursion in ConfigDocs worked, but I also wished if it actually broke coz that made more sense to me."

So  its not by surprise, we end up discussing ConfigDocs again :) I looked at your change in PR (not full PR yet), and it looks super-cool to me. And yes, that redundancy should be avoided, and good that we managed to do it.!

Thanks for your work.

afsalthaj:
@vigoo  This is very minor thing. The PR comes with a `to` functionality which is a real win.
I couldn't figure out a better heading for this PR. It will be good to specify something along those lines as well.

core/src/main/scala-2/zio/config/TupleConversion.scala
```diff
@@ -0,0 +1,86 @@
+package zio.config
+
+import scala.language.experimental.macros
+import scala.reflect.macros.whitebox
+
+trait TupleConversion[A, B] {
+  def to(a: A): B
+  def from(b: B): A
+}
+object TupleConversion {
```
afsalthaj, line 10: Lets have a space between 9 and 10

core/src/main/scala-2/zio/config/TupleConversion.scala
```diff
@@ -0,0 +1,86 @@
...
+  def apply[Prod, Tup]: TupleConversion[Prod, Tup] = macro genTupleConversion[Prod, Tup]
```
afsalthaj, line 11: I personally prefer just `A` and `B` over `Prod` and `Tup`.

core/src/main/scala-2/zio/config/TupleConversion.scala
```diff
@@ -0,0 +1,86 @@
...
+        // Special case: single parameter products
+
+        val typ = singleParam.typeSignatureIn(prodTpe).finalResultType
+
+        q"""new _root_.zio.config.TupleConversion[$prodTpe, $typ] {
+                override def to(a: $prodTpe): $typ = a.${singleParam.name.toTermName}                
+                override def from(b: $typ): $prodTpe =
+                    new ${prodTpe}(b)
+        }
+        """
```
afsalthaj, line 34: Couldn't `case List(params)` handle single parameter products as well?

core/src/main/scala-2/zio/config/TupleConversion.scala
```diff
@@ -0,0 +1,86 @@
...
+
+      case List(params) =>
+        // Generic case: n > 1 parameters
+
```
afsalthaj, line 38: Should we need an extra line here?

core/src/main/scala-2/zio/config/TupleConversion.scala
```diff
@@ -0,0 +1,86 @@
...
+                override def from(b: $tup): $prodTpe =
+                new ${prodTpe}(..$unpackers)
+            }
+            """
+      case Nil =>
+        // Special case: zero parameter products
+        q"""new _root_.zio.config.TupleConversion[$prodTpe, Unit] {
+                override def to(a: $prodTpe): Unit = ()
+                override def from(b: Unit): $prodTpe =
+                    new $prodTpe()
```
afsalthaj, line 67: Interesting. General thought:
How is zio-config going to behave for something like:
```scala

.transformOrFail(_ => ZeroCaseClass.apply, ZeroCaseClass.unapply)

```

core/src/main/scala-2/zio/config/TupleConversion.scala
```diff
@@ -0,0 +1,86 @@
...
+          params.indices.map { idx =>
+            val accessor = TermName(s"_${idx + 1}")
+            q"b.$accessor"
+          }
+
+        q"""new _root_.zio.config.TupleConversion[$prodTpe, $tup] {
+                override def to(a: $prodTpe): $tup =
+                (..$packers)
+                override def from(b: $tup): $prodTpe =
+                new ${prodTpe}(..$unpackers)
```
afsalthaj, line 59: Should we need a `new` here?

core/src/main/scala-2/zio/config/TupleConversion.scala
```diff
@@ -0,0 +1,86 @@
...
+                override def to(a: $prodTpe): Unit = ()
+                override def from(b: Unit): $prodTpe =
+                    new $prodTpe()
+            }
+        """
+      case _ =>
+        c.abort(
+          c.enclosingPosition,
+          s"Type ${prodTpe.typeSymbol} has multiple parameter lists which is currently not supported"
+        )
```
afsalthaj, line 74: I didn't really get this part. Could you explain a bit more?

core/src/main/scala-2/zio/config/TupleConversion.scala
```diff
@@ -0,0 +1,86 @@
...
+
+    c.Expr(result)
+  }
+}
+trait ImplicitTupleConversion {
+  implicit def autoTupleConversion[Prod <: Product, Tup]: TupleConversion[Prod, Tup] =
+    macro TupleConversion.genTupleConversion[Prod, Tup]
+
+  def autoTupleConversion1[Prod, A](implicit ev: TupleConversion[Prod, Tuple1[A]]): TupleConversion[Prod, A] =
+    ??? // defined here so it can be used in explicit imports when cross-compiling
```
afsalthaj, line 85: Could you explain a bit more on this as well? :)

core/src/main/scala-3/zio/config/TupleConversion.scala
```diff
@@ -0,0 +1,27 @@
...
+            def from(b: m.MirroredElemTypes): Prod = m.fromProduct(b)
+        }
+
+    inline given autoTupleConversion1[Prod <: Product, A](using c: TupleConversion[Prod, Tuple1[A]]): TupleConversion[Prod, A] =
+        new TupleConversion[Prod, A] {
+            def to(a: Prod): A = {
+                val Tuple1(v) = c.to(a)
+                v
+            }
+            def from(b: A): Prod = c.from(Tuple1(b))
```
afsalthaj, line 25: Scala3 💯

core/src/main/scala/zio/config/ConfigDescriptorModule.scala
```diff
@@ -98,6 +98,12 @@ trait ConfigDescriptorModule extends ConfigSourceModule { module =>
           )(Right(_))
       )
 
+    def to[B <: Product](implicit conv: TupleConversion[B, A]): ConfigDescriptor[B] =
```
afsalthaj, line 101: We need some API docs.

core/src/main/scala/zio/config/ConfigDocsModule.scala
```diff
@@ -639,27 +636,17 @@ trait ConfigDocsModule extends WriteModule {
...
+            path,
+            loopTo(
+              source.names ++ sources,
+              List.empty,
+              c,
+              Some(path),
+              alreadySeen
+            ),
+            descriptions
+          )
```
afsalthaj, line 649: 👍  I think this change makes sense!

core/src/main/scala/zio/config/ConfigDocsModule.scala
```diff
@@ -529,6 +533,7 @@ trait ConfigDocsModule extends WriteModule {
       case object AllOf         extends Format
       case object NotApplicable extends Format
       case object Recursion     extends Format
+      case object RecursionList extends Format
```
afsalthaj, line 536: I see 👍

core/src/main/scala/zio/config/ProductBuilder.scala
```diff
@@ -22,6 +22,12 @@ private[config] trait ProductBuilder[F[_], A, B] { self =>
   def apply[C](f: (A, B) => C, g: C => Option[(A, B)]): F[C] =
     a.zip(b).xmapEither({ case (aa, bb) => Right(f(aa, bb)) }, liftWrite(g))
 
+  def to[C](implicit conv: TupleConversion[C, (A, B)]): F[C] =
+    a.zip(b).xmapEither(
+      { case (aa, bb) => Right(conv.from((aa, bb) ))},
```
afsalthaj, line 27: Minor. I know I disabled scalafmt intentionally for this file.

```scala
a.zip(b).xmapEither(
 { case (aa, bb) => case (aa, bb) => Right(conv.from((aa, bb))) },

``` 
  


```

File: core/src/test/scala/zio/config/CommandLineListAccumulationTest.scala [hunk 15]
```diff
@@ -27,7 +27,7 @@ object CommandLineListAccumulationTest extends DefaultRunnableSpec {
 
   object SomeConfig {
     val descriptor: ConfigDescriptor[SomeConfig] =
-      list("ints")(int)(SomeConfig.apply, SomeConfig.unapply)
+      list("ints")(int).to[SomeConfig]
```
afsalthaj, line 30: Perfect. I hope it's all backward compatible.

core/src/test/scala/zio/config/CoproductTest.scala
```diff
@@ -119,7 +120,7 @@ object CoproductTestUtils {
     case class I(i: Double)           extends Z
   }
 
-  final case class Ldap(value: String) extends AnyVal
+  final case class Ldap(value: String)
```
afsalthaj, line 123: I can't remember, fo we have a special consideration for AnyVal in magnolia. I don't remember myself doing it though. Anyway cool!

core/src/test/scala/zio/config/ListsCornerCasesTest.scala
```diff
@@ -7,279 +7,277 @@ import zio.config.ReadError.{ FormatError, ListErrors, ZipErrors }
...
-              )
-            )
-
-          assert(res)(isLeft(equalTo(expected)))
-        }
-      )
+        assert(res)(isLeft(equalTo(expected)))
+      }
     )
+}
```
afsalthaj, line 283: Difficulty knowing what exactly changed.. But I hope its all good.

core/src/test/scala/zio/config/ProductBuilderTest.scala
```diff
@@ -27,8 +29,8 @@ object ProductBuilderTest
         },
         testM("combine 22 for tupled") {
           checkM(genS22) { p =>
-            val tuple =
-              S22.unapply(p).get
+            val conv = implicitly[TupleConversion[S22, (Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, Int)]]
```
afsalthaj, line 32: Should we add a new test case for TupleConversion and keep the old one as it is?

core/src/test/scala/zio/config/ReadWriteRoundtripTest.scala
```diff
@@ -142,20 +143,20 @@ object ReadWriteRoundtripTestUtils {
...
-  val cId             = string("kId")(Id.apply, Id.unapply)
-  val cId2            = string("kId2")(Id.apply, Id.unapply)
-  val cDataItem       = (cId2.optional |@| int("kDiCount"))(DataItem.apply, DataItem.unapply)
-  val cDbUrl          = string("kDbUrl")(DbUrl.apply, DbUrl.unapply)
-  val cEnterpriseAuth = (cId |@| cDbUrl)(EnterpriseAuth.apply, EnterpriseAuth.unapply)
+  val cId             = string("kId").to[Id]
+  val cId2            = string("kId2").to[Id]
+  val cDataItem       = (cId2.optional |@| int("kDiCount")).to[DataItem]
+  val cDbUrl          = string("kDbUrl").to[DbUrl]
+  val cEnterpriseAuth = (cId |@| cDbUrl).to[EnterpriseAuth]
```
afsalthaj, line 152: I hope things are backward compatible here.

project/BuildHelper.scala
```diff
@@ -73,19 +84,72 @@ object BuildHelper {
...
+    crossScalaVersions += Scala3,
+    scalacOptions ++= {
+      if (isDotty.value)
+        Seq("-noindent")
+      else
+        Seq()
+    },
+    scalacOptions --= {
+      if (isDotty.value)
+        Seq("-Xfatal-warnings")
```
afsalthaj, line 134: Do we need this ?

core/src/main/scala-2/zio/config/TupleConversion.scala
vigoo: I prefer meaningful type parameter names. If it's the length, what about `P` and `T`? It is important that `Prod` must be the case class and `Tup` the tuple.

core/src/main/scala-2/zio/config/TupleConversion.scala
vigoo: The initial version called the companion object's apply instead but it did not work in all cases. In some tests the companion object was `null`, maybe because of a similar issue than the one found in Scala 3, with the macro being invoked from base constructor invocation. This way it is safe and always works.

core/src/main/scala-2/zio/config/TupleConversion.scala
vigoo: You can't use it for types with multiple parameter lists, like `class X(a: Int, b: String)(c: Double)`

core/src/main/scala-2/zio/config/TupleConversion.scala
vigoo: Looks like it was not needed in the end for the tests, but at one point I thought I need to explicitly import the implicit in Scala3 where it is not `???` but does real work. And because in Scala 3 you can't import the givens automatically with `_` just by name, the code would not be backward source compatible if we don't have an equivalent symbol defined for Scala2 too.

core/src/test/scala/zio/config/CoproductTest.scala
vigoo: I had to remove `AnyVal` because the Scala3 product mirror does not work with it. People not interested in Scala3 migration can still use it, with the Scala2 macro it works.

core/src/test/scala/zio/config/ListsCornerCasesTest.scala
vigoo: Actually I think I mostly only changed the apply/unapply pairs to `to` but I guess that made scalafmt reformat things a bit as the lines became shorter..

core/src/test/scala/zio/config/ProductBuilderTest.scala
vigoo: If we do that it has to be moved to a Scala2-only source set as it won't compile on Scala3

project/BuildHelper.scala
vigoo: No because it's not in `stdOptions`

core/src/test/scala/zio/config/ListsCornerCasesTest.scala
vigoo: Actually if you "hide whitespace changes" on the github ui, then it shows it nicely. Not sure why, I guess it was scalafmt'd before too..

core/src/main/scala-2/zio/config/TupleConversion.scala
vigoo: For n>1 we assume that `Tup` is a tuple and deconstruct it to the case class parameters. For n=1 we assume its not a tuple and use the value directly. This makes it more usable in practice as you can write `string("foo").to[X]` for example. 
In the Scala3 version there is an additional implicit that automatically converts `A` to `Tuple1[A]` first.

core/src/main/scala-2/zio/config/TupleConversion.scala
afsalthaj: `P` and `T` would be ok. Considering we used `K` and `V` for key and value across the app, and stays consistent with that approach.

core/src/main/scala-2/zio/config/TupleConversion.scala
afsalthaj: Ah. Ofcourse. Makes sense.

core/src/main/scala-2/zio/config/TupleConversion.scala
afsalthaj: Sure. Makes sense.

core/src/main/scala-2/zio/config/TupleConversion.scala
afsalthaj: I see. Fair enough. We don't need to support it.
Have we tested `case class X[A](a: A, b: B)` ?

core/src/main/scala-2/zio/config/TupleConversion.scala
afsalthaj: I see.

core/src/test/scala/zio/config/ProductBuilderTest.scala
afsalthaj: Right.. Nope this is fine.

afsalthaj:
@vigoo  Are you making any more changes? If not, happy to approve and merge :)
Once again, great contribution!

afsalthaj (approved):

PR #535, opened 2021-05-12
Fix website 2

sideeffffect:
/cc @mijicd

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Cache scala dependencies
         uses: coursier/cache-action@v6
       - name: Check Document Generation
-        run: sbt docs/mdoc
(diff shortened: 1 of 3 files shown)


sideeffffect:
@mijicd When we test in CI only `docs/mdoc`, but not `docs/docusaurusCreateSite`, we do not prevent failures in master when trying to deploy the website :crying_cat_face:

mijicd (dismissed):

sideeffffect:
green :tada:

mijicd (approved):

sideeffffect:
We're getting this. Any idea how to fix that?
```
ERROR: Permission to zio/zio-config.git denied to deploy key
```
https://github.com/zio/zio-config/runs/2565386218?check_suite_focus=true#step:4:372

mijicd:
@softinio is the keeper of the keys, so he's the only one that can help here :)

PR #574, opened 2021-07-04
Scala3 auto derivation with zero dependency

afsalthaj:
Implements Scala3 auto-derivation with zero dependency. 
With scala3 auto derivation, we will try and avoid the need of DeriveConfigDescriptor trait and its heavy weight design.
There will not be SealedTraitStrategy. Instead we pushed these functionalities back into existing annotations (compile time driven) making it further easy to work with library.

image

Following limitations most likely to exist during the first release.

1. enums do possess difference in behaviour with respect to annotations. Hence enums will not be guaranteed to work with zio-config. While a general test do pass, getting into the details break property based tests (especially writing back and annotations). Hence we recommend users to go on with `sealed trait ` until a future release of zio-config fixes enum behaviour.

2. Recursive implicit resolutions may fail.  Hence we recommend scala3 users, use manual config derivation if you have recursive config. This doesn't mean you need to get away from auto-derivation completely. For instance, see this example:

https://github.com/zio/zio-config/blob/scala3_auto_derivation/magnolia/shared/src/main/scala-dotty/zio/config/magnolia/examples/RecursiveExample.scala

TODO:

- [x] Handle default values
- [ ] Handle pure-config's way of handling coproducts using `label(name = type)` annotation
- [x] Testing with relatively large/complex config is in progress. There are a few issues with respect to dotty implicit resolutions. The ETA has been changed from 15th to 31st.

ETA for this PR to get merged: July 31st.


magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Constant.scala
```diff
@@ -0,0 +1,21 @@
+package zio.config.magnolia
+
+import zio.config._
+
+/**
+ * A copy of zio-config-derivation util, to be able to map a constant value
+ * to case class that has the same name as that of the constant value, with zero
+ * number of fields in it.
+ */
+case class Constant(value: String) extends PropertyType[String, String] {
```
guizmaii, line 10: missing final

magnolia/shared/src/test/scala/zio/config/magnolia/AutomaticConfigDescriptorSpec.scala
```diff
@@ -58,7 +58,8 @@ object AutomaticConfigTestUtils {
 
   final case class Aws(region: String, security: Credentials)
 
-  final case class DbUrl(value: String) extends AnyVal
+  // Scala3 didn't like final case class DbUrl(value: String) extends AnyVal
```
guizmaii, line 61: What do you mean by "didn't like"?

afsalthaj: Hey Jules, thanks for coming up here and review.
Yes, scala3 gave out a unreasonable error during implicit inline resolution when giving `extends AnyVal`. End of the day, zio-config never had any special behaviour for `AnyVal`. Users, who need a new type is encouraged to define custom `Derivation` than injecting more magic into the library.

image

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Constant.scala
afsalthaj: Yes, absolutely it should be final.
These will be added as soon as the PR moves from Draft to an active one.

afsalthaj:
Recursive typeclass derivation in scala3 may give a hard time. However, in zio-config we can have work arounds by mixing auto-derivation with manual-derivation. It is quite straight forward, and is given in the below example:
https://github.com/zio/zio-config/blob/scala3_auto_derivation/magnolia/shared/src/main/scala-dotty/zio/config/magnolia/RecursiveExample.scala

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Descriptor.scala
```diff
@@ -0,0 +1,165 @@
+package zio.config.magnolia
```
afsalthaj, line 1: Even if it gives a bit of performance drill down, move away from using Strings in arguments. Example: Use List[SubClassName] instead of List[string]

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Macros.scala
```diff
@@ -0,0 +1,38 @@
...
+
+    Expr.ofList {
+      tpe.typeSymbol.primaryConstructor.paramSymss.flatten.map { field =>
+        Expr(field.name) -> field.annotations.filter { a =>
+          a.tpe.typeSymbol.maybeOwner.isNoSymbol ||
+            a.tpe.typeSymbol.owner.fullName != "scala.annotation.internal"
+        }.map(_.asExpr.asInstanceOf[Expr[Any]])
+      }.filter(_._2.nonEmpty).map { (name, anns) => Expr.ofTuple(name, Expr.ofList(anns)) }
+    }
+end Macros
```
afsalthaj, line 38: field level macros to be added.
I couldn't find any other way than dealing with macros to implement annotation driven derivation.

magnolia/shared/src/test/scala/zio/config/magnolia/AutomaticConfigDescriptorSpec.scala
```diff
@@ -100,8 +101,10 @@ object AutomaticConfigTestUtils {
       port           <- Gen.anyInt
       amount         <- Gen.option(Gen.long(1, 100))
       quantity       <- Gen.either(Gen.long(5, 10), genAlpha)
-      default        <- Gen.option(Gen.anyInt)
-      anotherDefault <- Gen.option(Gen.boolean)
+      default        <-
+        Gen.anyInt.map(Option(_)) // FIXME: Scala3 needs to handle default value. Replace with Gen.option(Gen.anyInt)
+      anotherDefault <-
+        Gen.boolean.map(Option(_)) // FIXME: Scala3 needs to handle default value. Replace with Gen.option(Gen.boolean)
```
afsalthaj, line 107: To fix before merge.

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Descriptor.scala
```diff
@@ -0,0 +1,165 @@
...
+         Descriptor(descOfList.transform[T](f, g), fieldNames)
+
+  def castTo[T](a: Any): T =
+    a.asInstanceOf[T]
+
+  extension[E, A](e: Either[E, A]) {
+    def mapError(f: E => String) =
+      e.swap.map(f).swap
+  }
+}
```
afsalthaj, line 165: Casting involved can be avoided if we go via the approach similar to 
https://github.com/afsalthaj/scala3-generic/blob/master/core/src/main/scala/com/thaj/scala/three/generic/typeclasses/MapperF.scala

However, after dealing with fairly non-idiomatic scala3 way, if by any chance we end up casting again, the whole exercise will be futile, resulting in a complex implementation to exist for particularly no reason.

Example: The following cannot be implemented in scala3 without `asInstanceOf`


```scala
  def sequence(s: TupleMap[F, T]): F[TupleInversMap[F, T]] = ???
```

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Descriptor.scala
```diff
@@ -0,0 +1,165 @@
...
+import java.time.{Instant, LocalDate, LocalDateTime, LocalTime}
+import java.util.UUID
+import scala.concurrent.duration.{Duration => ScalaDuration}
+import scala.deriving._
+import scala.compiletime.{erasedValue, summonInline, constValue, summonFrom, constValueTuple}
+import scala.quoted
+import scala.util.Try
+
+final case class Descriptor[A](desc: ConfigDescriptor[A], keys: List[String] = Nil) {
+  def isObject: Boolean = keys.isEmpty
```
afsalthaj, line 18: Not a great idea. Remove `isObject`

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Descriptor.scala
```diff
@@ -0,0 +1,165 @@
...
+}
+
+object Descriptor {
+  def apply[A](implicit ev: Descriptor[A]) =
+    ev.desc
+
+  final case class SubClassName(name: String, fields: List[String]){
+    def isObject = fields.isEmpty
```
afsalthaj, line 26: Unused. Remove SubClassName

magnolia/shared/src/test/scala/zio/config/magnolia/AutomaticConfigDescriptorSpec.scala
afsalthaj: Il need a macros to summon values explicitly for AnyVal

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Descriptor.scala
afsalthaj: Done

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Constant.scala
```diff
@@ -0,0 +1,21 @@
...
+    if (propertyValue == value) Right(value)
+    else Left(PropertyType.PropertyReadError(propertyValue, s"constant string '$value'"))
+
+  def write(a: String): String = a
+}
+
+object Constant {
+  def mk(value: String): ConfigDescriptor[String] =
+    ConfigDescriptorAdt.Source(ConfigSource.empty, Constant(value)) ?? s"constant string '$value'"
+}
```
afsalthaj, line 21: Add a new line here

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Descriptor.scala
```diff
@@ -0,0 +1,197 @@
...
+
+final case class Descriptor[A](desc: ConfigDescriptor[A], keys: List[FieldName] = Nil)
+
+object Descriptor {
+  def apply[A](implicit ev: Descriptor[A]) =
+    ev.desc
+
+  final case class FieldName(names: List[String])
+  final case class SubClassName(originalName: String, alternativeNames: List[String])
+  final case class ParentName(originalName: String, alternativeNames: List[String])
```
afsalthaj, line 26: Same approach should be taken for scala-2.12 magnolia

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Descriptor.scala
```diff
@@ -0,0 +1,185 @@
...
+  def mergeAllProducts[T](
+    allDescs: => List[Descriptor[T]],
+  ): Descriptor[T] =
+    Descriptor(
+      allDescs
+        .map(desc =>
+          desc.metadata match {
+            case Some((productName, fields))  if (fields.nonEmpty) =>
+              (productName.alternativeNames :+ productName.originalName)
+                .map(n => nested(n)(desc.desc)).reduce(_ orElse _)
```
afsalthaj, line 134: define 

```scala
def tryAllKeys[A](keys: List[String], desc: Descriptor[A]) = ???

```

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Descriptor.scala
```diff
@@ -0,0 +1,185 @@
...
+    tryAllParentPaths(coproductName, desc)
+
+  def tryAllParentPaths[T](
+    coproductName: CoproductName,
+    desc: Descriptor[T]
+  ) =
+    if coproductName.alternativeNames.isEmpty then
+      desc
+    else
+      Descriptor(coproductName.alternativeNames.map(n => nested(n)(desc.desc)).reduce(_ orElse _))
```
afsalthaj, line 123: use tryAllKeys defined below

afsalthaj:
It is baffling why `sym.tree` actually fails when trying dotty macros.
This is required for identifying `default` values. The delay in this PR to be `non-draft` is due to this error

image

afsalthaj:
Good news. Solved the issue mentioned above.

magnolia/shared/src/test/scala/zio/config/magnolia/AutomaticConfigDescriptorSpec.scala
```diff
@@ -76,15 +76,15 @@ object AutomaticConfigTestUtils {
...
   private val genPrice: Gen[Random, Price]       = Gen.oneOf(genPriceDescription, genCurrency)
 
-  private val genToken       = genNonEmptyString(5).map(Token)
-  private val genPassword    = genNonEmptyString(5).map(Password)
+  private val genToken       = genNonEmptyString(5).map(Token.apply)
+  private val genPassword    = genNonEmptyString(5).map(Password.apply)
   private val genCredentials = Gen.oneOf(genToken, genPassword)
 
-  private val genDbUrl = genNonEmptyString(5).map(DbUrl)
+  private val genDbUrl = genNonEmptyString(5).map(DbUrl.apply)
```
afsalthaj, line 87: Only to remove warnings in scala3.

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Descriptor.scala
```diff
@@ -0,0 +1,247 @@
+package zio.config.magnolia
+
```
afsalthaj, line 2: A few code refactoring such as separate functions for `product` and `sum` is reverted due to dotty errors.

afsalthaj:
@jdegoes  Could you review this?

afsalthaj:
Support pure-config labels can be in a follow up commit.
We can add specific tests for macros in the same commit.

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/Macros.scala
```diff
@@ -0,0 +1,62 @@
...
+    import quotes.reflect.*
+
+    val tpe = TypeRepr.of[T]
+
+    Expr.ofList {
+      tpe.typeSymbol.annotations.filter { a => {
+         a.tpe.typeSymbol.fullName == ownerName
+      }}.map(_.asExpr.asInstanceOf[Expr[A]])
+    }
+  }
```
afsalthaj, line 48: Remove `{}` to stay consistent

jdegoes (approved):

jdegoes (approved):

PR #583, opened 2021-07-26
Add `refineType` function that accepts a fully formed refined type

### LINKED ISSUE #571: Add `refineType` function that accepts a fully formed refined type
Example:

Currently refine module can accept predicates represented by types.

Examples:
```scala

refine[NonEmpty](string("ENV"))

// Equivalent to 

refine[String, NonEmpty]("ENV")

```

Another example:

```scala
refine[NonEmpty](list("regions")(string))

```

Another example:

```scala

refine[Size[Greater[W.`2`.T]]](list("regions")(string))

```

and so on, and it composes quite well.

However, there are situtations where users expect it to work by giving a fully formed refined types.

Example:
```scala
type NonEmptyString = String Refined NonEmpty

refine[NonEmtpyString]("ENV")

// As of now the above code doesn't compile coz NonEmtpyString isn't a predicate anymore. Hence the solution with version 1.0.6 is 

refine[String, NonEmpty]("ENV") // or refine[NonEmpty](string("ENV"))
```

Add a new `refine` function that can accept a fully formed type as mentioned in the last example.

afsalthaj:
Fixes  #571 .

This PR adds a new method to refined package allowing users to pass fully formed refined-types.

That is;

```scala    
  type NonEmptyString = String Refined NonEmpty
  
  val config: ConfigDesriptor[NonEmptyString] = 
    refineType[NonEmtpyString]("ENV")


  // Currently, the only way to interact is to specify the predicate directly
  
  refine[String, NonEmpty]("ENV") 
  
  // or 
  
  refine[NonEmpty](ConfigDescriptor.string("ENV"))
  

```


refined/shared/src/main/scala/zio/config/refined/PartialRefinedPath.scala
```diff
@@ -0,0 +1,25 @@
...
+    d: Descriptor[A]
+  ): ConfigDescriptor[F[A, P]] =
+    nested(path)(
+      d.desc
+        .transformOrFail[F[A, P]](
+          t => rt.refine[P](t).map(_.asInstanceOf[F[A, P]]), // =:= ev.apply doesn't work with 2.12.13
+          rf => Right(rt.unwrap(rf))
+        )
+    )
+}
```
afsalthaj, line 25: `asInstanceOf` is used due to the following compiler error.

https://github.com/zio/zio-config/runs/3158656513

PR #612, opened 2021-09-05
refactor zip case for write

yangzai:
`write` should not produce new errors on `Zip` as the `merge` method essentially returns a `NonEmptyList`.

On another note, `TransformOrFail` is the only other case that produces new errors. In general, I don't think `write` should fail but  `TransformOrFail` allows for failures for transformation in both directions.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -167,7 +167,6 @@ lazy val zioConfig = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(
     libraryDependencies ++= Seq(
       "dev.zio"                %% "zio"                     % zioVersion,
-      "dev.zio"                 %% "zio-prelude"            % "1.0.0-RC6",
       "org.scala-lang.modules" %% "scala-collection-compat" % "2.5.0",
       "dev.zio"                %% "zio-test"                % zioVersion % Test
     ),
--- a/build.sbt
+++ b/build.sbt
@@ -167,6 +167,7 @@ lazy val zioConfig = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(
     libraryDependencies ++= Seq(
       "dev.zio"                %% "zio"                     % zioVersion,
+      "dev.zio"                 %% "zio-prelude"            % "1.0.0-RC6",
       "org.scala-lang.modules" %% "scala-collection-compat" % "2.5.0",
       "dev.zio"                %% "zio-test"                % zioVersion % Test
     ),
--- a/core/shared/src/main/scala/zio/config/PropertyTree.scala
+++ b/core/shared/src/main/scala/zio/config/PropertyTree.scala
@@ -1,6 +1,7 @@
 package zio.config
 
 import com.github.ghik.silencer.silent
+import zio.prelude.NonEmptyList
 
 import scala.annotation.tailrec
 import scala.collection.immutable.Nil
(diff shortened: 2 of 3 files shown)


build.sbt
```diff
@@ -160,6 +160,7 @@ lazy val zioConfig = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .settings(
     libraryDependencies ++= Seq(
       "dev.zio"                 %% "zio"                     % zioVersion,
+      "dev.zio"                 %% "zio-prelude"             % "1.0.0-RC6",
```
yangzai, line 163: I'm not sure if it should depend on `zio-prelude` for the `NEL` but alternatively we could still get the unsafe `head` from the the resultant list and replace `NEL` with a property check on `merge`.

core/shared/src/main/scala/zio/config/WriteModule.scala
```diff
@@ -69,11 +69,9 @@ private[config] trait WriteModule extends ConfigDescriptorModule {
           val rightResult = go(cd.right, tuple._2)
 
           for {
-            left   <- leftResult
-            right  <- rightResult
-            merged  = left.merge(right)
-            result <- merged.headOption.toRight("Failed to write the config back to property tree, at zip node")
-          } yield result
+            left  <- leftResult
+            right <- rightResult
```
afsalthaj, line 73: I can’t remember the top of my head while I was writing this loop. But just by looking at the change, I think it makes sense.

core/shared/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -116,12 +117,12 @@ sealed trait PropertyTree[+K, +V] { self =>
...
    * @return
    */
-  final def merge[K1 >: K, V1 >: V](that: PropertyTree[K1, V1]): List[PropertyTree[K1, V1]] =
+  final def merge[K1 >: K, V1 >: V](that: PropertyTree[K1, V1]): NonEmptyList[PropertyTree[K1, V1]] =
     (self, that) match {
-      case (Sequence(l), Sequence(r))           => singleton(Sequence(l ++ r))
+      case (Sequence(l), Sequence(r))           => NonEmptyList(Sequence(l ++ r))
       case (l: Record[K, V], r: Record[K1, V1]) =>
         (l.value.keySet ++ r.value.keySet)
-          .foldLeft(List[Map[K1, PropertyTree[K1, V1]]](Map.empty)) { case (acc, k) =>
```
afsalthaj, line 124: Hmm. Just for NonEmptyList, we are not yet in a position to depend on Zio-prelude. A temporary set up for NonEmptyList in prelude package is a nice idea. In future, I can see zio-config depending on zio-prelude though.

afsalthaj:
TransformOrFail allowing an errorneous output during write is, I think, not a luxury. In a way we can think someone can populate a invalid product and ask it to write back a valid config string. But that all said, a certain level of compromise for ergonomic usage is ok.

core/shared/src/main/scala/zio/config/PropertyTree.scala
yangzai: @afsalthaj I could implement a minimal `NEL` based of `zio-prelude` for now. Let me know where I should place the new `NEL`.

yangzai: @afsalthaj I've just realised that there's a `NonEmptyChunk` that does not depend on `zio-prelude`. I'm not sure if it affects perfomance switch from `List` to `Chunk`.

afsalthaj: yeah good point. It’s always good to move from List to Chunk.

yangzai: @afsalthaj I've switched to `NonEmptyChunk` and dropped the `zio-prelude` dependency.

afsalthaj: Fantastic. I am facing some scala3 issues in master branch. So might delay the release.

afsalthaj:
This is such a deep dive into library and thanks a lot of taking your time @yangzai

afsalthaj (approved):

PR #613, opened 2021-09-08
add pureconfig module and typesafe configvalue descriptor

yangzai:
Derives `ConfigDescriptor` from Pureconfig `ConfigConvert`.

This requires a `ConfigValue` source descriptor as `ConfigConvert` provides the necessary transformation to and from `ConfigValue`.

Will add more test cases and docs if the general concept is OK.


typesafe/shared/src/main/scala/zio/config/typesafe/TypesafePropertyType.scala
```diff
@@ -0,0 +1,21 @@
...
+import zio.config._
+
+import scala.util.Try
+
+object TypesafePropertyType {
+  import PropertyType.PropertyReadError
+
+  object ConfigValuePropertyType extends PropertyType[String, ConfigValue] {
+    override def read(propertyValue: String): Either[PropertyReadError[String], ConfigValue] =
+      Try(ConfigFactory.parseString(propertyValue).resolve).map(_.root)
```
yangzai, line 13: I couldn't find an accessible `ConfigValue` parser so this is kind of a hack. First we try parsing the string as a `Config` and if it succeeds we extract the `ConfigObject` within the `Config`. If we fail we prepend a key so it essentially gets wrapped into an object and we try parsing it again. If it succeeds we extract the value with the key.

The reason we don't skip the first parse is that because a hocon object might not be wrapped within parenthesis and prepend a key to it would produce an invalid string for the second parse.

afsalthaj:
I will take a look

yangzai:
@afsalthaj I fixed some CI issues and it needs approval again

afsalthaj (approved):

afsalthaj:
@yangzai  you might need to wait for this PR to get merged.
https://github.com/zio/zio-config/pull/617

afsalthaj:
While almost all of the pure-config way of derivation is supported in zio-config, I believe this change is still invaluable especially for quick migrations. Once again thanks for the PR, and I will be merging this as soon as the build is ready (I have updated again), and schedule for the next release.

typesafe/shared/src/main/scala/zio/config/typesafe/TypesafeConfigDescriptor.scala
```diff
@@ -0,0 +1,9 @@
+package zio.config.typesafe
+
+import com.typesafe.config._
+import zio.config._
+
+object TypesafeConfigDescriptor {
```
afsalthaj, line 6: Hey could we rename this to PureConfigDescriptor ?

typesafe/shared/src/main/scala/zio/config/typesafe/TypesafePropertyType.scala
```diff
@@ -0,0 +1,27 @@
+package zio.config.typesafe
+
+import com.github.ghik.silencer.silent
+import com.typesafe.config._
+import zio.config._
+
+import scala.util.Try
+
+@silent("Unused import")
+object TypesafePropertyType {
```
afsalthaj, line 10: Could we rename to ConfigValuePropertyType?

afsalthaj:
Sorry, I realised later, that a few renaming is important, just to stop confusing users between what's provided in zio-config-typesafe and this module.

typesafe/shared/src/main/scala/zio/config/typesafe/TypesafePropertyType.scala
yangzai: @afsalthaj but the inner object is already `ConfigValuePropertyType`. We could drop the outer `TypesafePropertyType`.

typesafe/shared/src/main/scala/zio/config/typesafe/TypesafeConfigDescriptor.scala
yangzai: @afsalthaj how about we just move the inner `configValueConfigDescriptor` into package and drop `TypesafeConfigDescriptor`?

typesafe/shared/src/main/scala/zio/config/typesafe/TypesafePropertyType.scala
afsalthaj: Makes sense.

afsalthaj:
Waiting for this change to finally merge for our next release.

afsalthaj (approved):

PR #618, opened 2021-09-13
fixed silencer version

sviezypan:
value Scala213 resolved to "2.13" instead of "2.13.6" and that failed to download silencer-lib dependency, which caused that newest release "1.0.7" failed for scala 3


afsalthaj:
But I guess scala3 is still failing in this PR?

sviezypan:
let me try to fix this

afsalthaj:
https://github.com/lampepfl/dotty/issues/13406

The error seem to happen when we summon a product-mirror for a 2.13-compiled artifact.
Not sure how much related this is to our error.

afsalthaj:
Any luck on this?

sviezypan:
No, I am stuck. :( Locally sbt ++3.0.2! testDotty and publishLocal commands are working. So I didn't even reproduce this.
Instead I focused on adding scala 3 support to the remaining modules and while in "examples" module it was mostly just changing explicit unapply calls (which doesn't return Option in scala3),  "shapeless" module is really difficult for me to migrate.
If this error is unrelated to the fact that some of the modules are not scala 3 compatible yet, then I don't know how to proceed.

afsalthaj:
Hmm. Yea to be honest, personally I don think we need to support zio-Config-shapeless for scala3. The scala3 auto derivation is anyway independent of dependencies like magnolia and shapeless. So let’s start calling zio-Config-shapeless deprecated and allow it to disappear soon .

sviezypan:
@afsalthaj so I was able to reproduce only with jdk8. As you wrote that magnolia is not needed for scala3 I tried to removed it from testDotty command and scala3projects and now build passes. Let me know if thats something that makes sense. Other than that I don't know how to solve this.

afsalthaj:
@sviezypan  Hmm. As far as it test this code https://github.com/zio/zio-config/tree/master/magnolia/shared/src/main/scala-dotty/zio/config/magnolia, all good

Although the name is zio-config-magnolia, in scala3, its implemented without magnolia

afsalthaj:
In other words, these tests https://github.com/zio/zio-config/tree/master/magnolia/shared/src/test/scala/zio/config/magnolia are only for dotty.

build.sbt
```diff
@@ -74,7 +74,7 @@ addCommandAlias(
 )
 addCommandAlias(
   "testDotty",
-  ";zioConfigJVM/test;zioConfigTypesafeJVM/test;zioConfigDerivationJVM/test;zioConfigYamlJVM/test;zioConfigMagnoliaJVM/test"
```
afsalthaj, line 77: I think, hmm, we need to run scala3 tests.
This will give you a hint why so

https://github.com/zio/zio-config/blob/95c2f0b49813637a93aa0129753f7b09da016ae3/build.sbt#L85

afsalthaj:
```scala

lazy val magnoliaDependencies =
  libraryDependencies ++= {
    if (scalaBinaryVersion.value == "2.11" || scalaVersion.value == ScalaDotty) Seq.empty // Just to make IntelliJ happy
    else {
      Seq(
        "com.propensive" %% "magnolia"      % magnoliaVersion,
        "org.scala-lang"  % "scala-reflect" % scalaVersion.value
      )
    }
  }

```

afsalthaj:
For now, I am just merge this and release a version. 
We will take it up later.

afsalthaj (approved):

afsalthaj:
Still having issue in the `publish` task in master branch

PR #633, opened 2021-10-02
Redesign (Simplify ConfigSource) and expand PropertyTree

afsalthaj:
## Features added

1. `ConfigSource.from` will not return `ZLayer` or `ZIO` anymore. This means users have more easier composability with `ConfigSource`.

```scala

  val source = ConfigSource.fromMap(Map("a" -> "b")) orElse TypesafeConfigSource.fromHoconFile(file)
  read(config from source)

```

2. Users are able to peek into a subtree of a source and make it `ConfigSource`, and thereby reduce redundant case class configurations, especially usefully when dealing with auto-derivations where your case class is not corresponding to the original source.

```scala
 {
    "a" : {
        "b" : {
            "c" : [
                {
                    "x"  : 1
                    "y" : 2
                }
                 {
                    "x" : 3
                    "y" : 4
                }
             ]
         }
     }
 }       


final case class ShortConfig(x: Int, y: Int)

read(descriptor[ShortConfig] from TypesafeConfigSource.fromHoconFile(file).at("a.b.c[0]"))

// returning ShortConfig(3, 4) 

```

3. Live updates of config is now possible. ConfigSource will pick the latest updates from the original source, for every `read` that you do, with the feature to do `source.cache` if you need to avoid live updates.
4. Consistent resource management in ConfigSource using `ZManaged`.

In short, `zio-config` is now more `zio-ish`, with `read` returning `IO[ReadError[K], A]` instead of `Either[ReadError[K], A]` and making full use of `ZManaged` and `ZIO`.

There are only very reasonable/minimal changes for the end users, where the only immediate change in their code would be around removing the need to flatMap over their previously `ZIOish` `ConfigSource`

5. Removing complex methods in `PropertyTree`. `PropertyTree` is now fairly light-weight.


afsalthaj:
Closing it for #635

PR #635, opened 2021-10-03
Simplify ReadModule, ConfigSource and PropertyTree

### LINKED ISSUE #658: Redesign ConfigSource, PropertyTree, ReadModule and Resource Handling in zio-config

afsalthaj:
Fixes #658 

## Features added

1. `ConfigSource.from` will **not** return `ZLayer` or `ZIO` anymore. This means we have more **easier composability** with `ConfigSource` regardless of the type of `ConfigSource`.

```scala

  val source = ConfigSource.fromMap(Map("a" -> "b")) orElse TypesafeConfigSource.fromHoconFile(file)
  read(config from source)

```

2. We can now **peek into a subtree** of a source and make it `ConfigSource`, and thereby reduce redundant case class configurations, especially usefully when dealing with auto-derivations where your case class is not corresponding to the original source.

```scala
 {
    "a" : {
        "b" : {
            "c" : [
                {
                    "x"  : 1
                    "y" : 2
                }
                 {
                    "x" : 3
                    "y" : 4
                }
             ]
         }
     }
 }       


final case class ShortConfig(x: Int, y: Int)

read(descriptor[ShortConfig] from TypesafeConfigSource.fromHoconFile(file).at("a.b.c[0]"))

// returning ShortConfig(3, 4) 

```

3. Memoization of config-source at various levels
   *  zio-config can memoize your config-source per read. Example: Reading 5 config values to fill a case class will involve
      talking to the actual source 5 times, unless we memoize ConfigSource. If there is a resource acquisition and release  involved in reading each config parameter, then acquire and release will happen 5 times as well. We can memoize ConfigSource by calling `.memoize`, and in this case the resource acquisition and release happens once, with an in-memory tree representing your config for further reads.
   * There is a second level of memoization where config-source can be read as a whole once and for all, and is reused in multiple reads. This memoization is done by calling `configSource.strictlyOnce` returning a ZIO effect indicating that the communication with the actual data source happens prior to the `reads`
      
4. Consistent resource management in ConfigSource using `ZManaged`.

5. Removing complex methods in `PropertyTree`. `PropertyTree` is now fairly light-weight.

6. Plus a whole bunch of cleaner design and logic in https://github.com/zio/zio-config/pull/636

In short, `zio-config` is now more `zio-ish`, with `read` returning `IO[ReadError[K], A]` instead of `Either[ReadError[K], A]` and making full use of `ZManaged` and `ZIO`.

There are only very reasonable/minimal changes for the end users, where the only immediate change in the code would be around removing the need to flatMap over their previously `ZIOish` `ConfigSource`. 

This can lead to a question of why a simple `map` source be `zio-ish`. However, we learned it through the hard way that the question is slightly irrelevant in a big picture of zio-config with almost 99% of config-sources actually need effect and resource handling. The cost of being pedantic about constant-sources was improper and inconsistent return types across all  config-sources. So, now, just about everything is `ZManaged` (`ZManaged[Any, ReadError[K], PropertyTreePath[K] => ZIO[Any, ReadError[K], PropertyTree[K, V]]` internally without exposing it to the user. Also, the solution to a `Map` or a constant `PropertyTree` having to be in  ZIO was nothing but simple lift to `ZIO` using `ZIO.succeed`. Also, note that users who used zio-config to be returning just `Either` without `zio` still had `zio` in their classpath.

## Make zio-config's ReadModule deal with PropertyTree until the recursion terminates.

 A more natural way of thinking about the `recursions` in `ReadModule` is, every node should return a `PropertyTree` capturing possible errors using simple `Either`. i.e, `Either[ReadError[K], PropertyTree[A]]`. This implies a `Sequence` node should just return `Either[ReadError[K], PropertyTree[List[A]]`. 
 
 Currently at every node of ConfigDescriptor in `read`, what we get in return is not  `PropertyTree` but a real `value` `A`. ie, `Either[ReadError[K], A]`. This might sound ok in first glance, but it comes with a few concerning logic inside `ReadModule` that can confuse developers. 
 
As an example of this logical fragileness,  in the current version, `ReadModule` has to keep track of `List[Step[K]]` as well as a `List[K]`, one to report errors, and the other to fetch it from `ConfigSource`. This indirectly also implies, at every node, especially `Sequence` and `Map`, we had to update the source. This requirement resulted in adding `ConfigSource` in `Map` and `Sequence` ConfigDescriptor`.

```scala
    // BEFORE
    case class Nested[A](source: ConfigSource, path: K, config: ConfigDescriptor[A]) extends ConfigDescriptor[A]
    case class Source[A](source: ConfigSource, propertyType: PropertyType[V, A]) extends ConfigDescriptor[A]
    case class Sequence[A](source: ConfigSource, config: ConfigDescriptor[A]) extends ConfigDescriptor[List[A]]
    case class DynamicMap[A](source: ConfigSource, config: ConfigDescriptor[A]) extends ConfigDescriptor[Map[K, A]]

```
 
 
While its ok for `ConfigDescriptor` to have the `Source` node, having the `Source` details explicitly in other nodes may confuse the future developers and maintainers of zio-config.  With this change, we now have

```scala

    // AFTER
    case class Nested[A](path: K, config: ConfigDescriptor[A]) extends ConfigDescriptor[A]
    case class Source[A](source: ConfigSource, propertyType: PropertyType[V, A]) extends ConfigDescriptor[A]
    case class Sequence[A](config: ConfigDescriptor[A]) extends ConfigDescriptor[List[A]]
    case class DynamicMap[A](config: ConfigDescriptor[A]) extends ConfigDescriptor[Map[K, A]]
```

From Day1, we tried to implement `read` through recursive logic within the `PropertyTree` context. However, we were taking our time, and we later accepted PRs that handle the read logic using `BEFORE` (above), which enabled us to release zio-config without further delay and get early feedbacks.

Another improvement is removal of `LeafForSequence` type and the type is avoided completely in the main API.
The concern is narrowed down to just `PropertyTree` and there is no usage of it in `ReadModule` making abstractions and methods look better orthogonal

- [x] Redesign core module and make all important tests including ones for recursive config pass.
- [x] Fix other modules
       - [x] Typesafe
       - [x] Core
       - [x] Magnolia
       - [x] scalaz
       - [x] cats
       - [x] refined
       - [x] gen
       - [x] typesafe-magnolia-tests
       - [x] pureconfig
       - [x] shapeless 
       - [x] examples
       - [x] yaml
       - [x] docs
       
- [x] Remove the conflict between `ConfigSource` case class and `ConfigSource` object in `ConfigSourceString` module. Currently, `ConfigSource_` is the name given temporarily to avoid conflict. 
- [x] Pull the caching to ZManaged level
- [x] Remove Step in ReadError as it is part of PropertyTreePath
- [x] Add test to prove memoizing
- [x] Merge https://github.com/zio/zio-config/pull/636


core/shared/src/main/scala/zio/config/ConfigDescriptorModule.scala
```diff
@@ -586,7 +586,7 @@ trait ConfigDescriptorModule extends ConfigSourceModule { module =>
      *       int("PORT") from configSource2.orElse(configSource1))(Config.apply, Config.unapply) from configSource2
      * }}}
      */
-    final def from(that: ConfigSource): ConfigDescriptor[A] =
+    final def from(that: ConfigSource_): ConfigDescriptor[A] =
```
afsalthaj, line 589: Renamed only temporarily.

core/shared/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -132,89 +212,16 @@ sealed trait PropertyTree[+K, +V] { self =>
             }
           }
           .map(v => PropertyTree.Record(v))
-      case (left, right) if left.isEmpty        => NonEmptyChunk(right)
-      case (left, right) if right.isEmpty       => NonEmptyChunk(left)
-      case (l, r)                               => NonEmptyChunk(l, r)
+      case (left, right) if left.isEmpty        => singleton(right)
```
afsalthaj, line 215: revert these changes

core/shared/src/main/scala/zio/config/ConfigSourceModule.scala
```diff
@@ -17,20 +17,24 @@ trait ConfigSourceModule extends KeyValueModule {
 
   case class ConfigSource(
     sourceNames: Set[ConfigSource.ConfigSourceName],
-    access: ZManaged[Any, ReadError[K], PropertyTreePath[K] => UIO[ZIO[Any, ReadError[K], PropertyTree[K, V]]]],
+    access: ZManaged[
+      Any,
+      Nothing,
+      ZManaged[Any, ReadError[K], PropertyTreePath[K] => ZIO[Any, ReadError[K], PropertyTree[K, V]]]
+    ],
```
afsalthaj, line 24: Note that squashing it to a single `ZManaged` layer results in non-memoization behaviour, which actually makes sense considering the type returned by `.memoize` on `ZManaged`.

core/shared/src/test/scala/zio/config/CommandLineListAccumulationTest.scala
```diff
@@ -11,7 +11,7 @@ object CommandLineListAccumulationTest extends DefaultRunnableSpec {
   def spec: Spec[TestEnvironment, TestFailure[Nothing], TestSuccess] =
     suite("Configuration of a list from multiple entries")(
       testM("Using single arg --key=value style") {
-        checkM(Gen.int(1, 10)) { count =>
+        checkM(Gen.int(1, 10).map(_ => 1)) { count =>
```
afsalthaj, line 14: ?

core/shared/src/main/scala/zio/config/ConfigSourceModule.scala
```diff
@@ -1,194 +1,184 @@
...
   }
 
-  protected object ConfigSourceFunctions extends ConfigSourceFunctions
-}
+  object ConfigSource {
+    type Managed[A]              = ZManaged[Any, ReadError[K], A]
+    type TreeReader              = PropertyTreePath[K] => ZIO[Any, ReadError[K], PropertyTree[K, V]]
+    type MemoizableManaged[A]    = ZManaged[Any, Nothing, ZManaged[Any, ReadError[K], A]]
+    type ManagedReader           = Managed[TreeReader]
+    type MemoizableManagedReader = MemoizableManaged[TreeReader]
```
afsalthaj, line 107: Make this private

core/shared/src/main/scala/zio/config/ConfigSourceModule.scala
```diff
@@ -1,194 +1,184 @@
...
 
+    val empty: ConfigSource =
+      Reader(
+        Set.empty,
+        ZManaged.succeed(ZManaged.succeed(_ => ZIO.succeed(PropertyTree.empty)))
+      )
+
+    case class OrElse(self: ConfigSource, that: ConfigSource) extends ConfigSource
+
+    case class Reader(
```
afsalthaj, line 123: sealed

core/shared/src/main/scala/zio/config/ConfigSourceModule.scala
```diff
@@ -391,42 +394,44 @@ trait ConfigSourceStringModule extends ConfigSourceModule {
...
+                              properties.load(inputStream)
+                              properties
+                            }
+
+              tree = getPropertyTreeFromProperties(
+                       properties,
+                       keyDelimiter,
+                       valueDelimiter,
+                       filterKeys
+                     )
```
afsalthaj, line 423: Move creation of trees up

core/shared/src/main/scala/zio/config/PropertyTree.scala
```diff
@@ -1,28 +1,158 @@
...
+      self match {
+        case Step.Index(n) => Step.Index(n)
+        case Step.Key(k)   => Step.Key(f(k.asInstanceOf[K1]))
+      }
+  }
+
+  object Step {
+    val pattern: Regex = """([a-zA-Z0-9 -@\-^-~]*)\[([0-9]*)\]""".r.anchored
+
+    def steps[K](s: String)(implicit IsString: String =:= K): Vector[Step[K]] =
```
afsalthaj, line 29:
```scala
<:<
```

core/shared/src/test/scala/zio/config/MemoizedSourceSpec.scala
```diff
@@ -0,0 +1,164 @@
...
+            v <- read(config from source)
+            r <- ZIO.succeed(resource.get)
+            c <- ZIO.succeed(configCount.get)
+          } yield (v, r, c)
+
+        assertM(effect)(equalTo(((s"v1_1_1", "v2_2_2"), 2, 2)))
+
+      },
+      testM(
+        "A non-memoized source runs a single resource acquisition when retrieving a list"
```
afsalthaj, line 43: description wrong.

jdegoes (approved):

PR #875, opened 2022-05-29
Suspend the creation and loading of Config in ZIO

sideeffffect:
This also bring the `TypesafeConfig`'s API in line with `TypesafeConfigSource` and thus makes the whole API more regular.


afsalthaj (approved):

afsalthaj:
Thanks @sideeffffect. Makes sense

PR #884, opened 2022-06-29
Removed used type parameters.

landlockedsurfer:
Hi!
The pull request resolves a minor issue removing some unused type parameters.
Hope this is useful!


afsalthaj:
Thanks for spotting and fixing it !

Issue #439, opened 2020-11-09
Support automatic derivation of value classes


leszekgruchala:
The main goal is to have automatic support for config descriptors that are value classes.

```scala
import zio.config.magnolia.DeriveConfigDescriptor.Descriptor

case class Port(value: Int) extends AnyVal
implicit val portDescriptor: Descriptor[Port] = Descriptor[Int].transform(Port, _.value)
```

Sample failing code https://scastie.scala-lang.org/9UDHZDYDRtmqiYybxGQwaA

remiguittaut:
with credits from https://discord.com/channels/629491597070827530/633028431000502273/775153057965015050.

```scala
import cats.evidence.As
import shapeless.ops.hlist.IsHCons
import shapeless.{Generic, HList, HNil}
import zio.config.magnolia.DeriveConfigDescriptor.Descriptor

implicit def getDescriptorForValueClasses[A, T <: HList, H](
    implicit I: A As AnyVal,
    G: Generic.Aux[A, T],
    IsH : IsHCons.Aux[T, H, HNil],
    CD: Descriptor[H]
 ): Descriptor[A] = CD.transform(value => G.to(value), t => G.from(t))
```

should work, but it would be preferable to avoid bringing in the shapeless dependency. a macro, with the inspiration from 
io.circe.generic.extras.deriveUnwrappedCodec would be great :)

afsalthaj:
The solution to support value classes will be baked into existing magnolia
derivation.

afsalthaj:
Thanks for reporting this issue. I will take a look at it

soujiro32167:
Looks like this was fixed in Magnolia: https://github.com/softwaremill/magnolia/issues/225

PR #909, opened 2022-08-02
Handle AnyVal #439

Esardes:


magnolia/shared/src/main/scala-2.12-2.13/zio/config/magnolia/Descriptor.scala
```diff
@@ -402,11 +402,14 @@ object Descriptor {
             acc orElse f(n)
           }
         case head :: tail =>
+          def nest(name: String)(unwrapped: ConfigDescriptor[Any]) =
```
afsalthaj, line 405: Do we have a test that confirms if the behaviour works? :)

afsalthaj:
Thanks @Esardes for the contribution. It looks good to me.
This is an important addition, there fore we need to test that works with all scala versions

magnolia/shared/src/main/scala-2.12-2.13/zio/config/magnolia/Descriptor.scala
Esardes: added a test, works for you?

afsalthaj (approved):

PR #917, opened 2022-08-12
Added pagination support

ithinkicancode:


aws/shared/src/main/scala/zio/config/aws/parameterstore/ParameterStoreConfigSource.scala
```diff
@@ -26,12 +27,36 @@ object ParameterStoreConfigSource {
...
-            .attempt(ssm.getParametersByPath(request).getParameters)
-            .map(_.asScala.toList)
-            .map { list =>
+          ZStream
+            .paginateZIO(
+              ZIO.attempt(ssm.getParametersByPath(request))
+            )(_.map { respponse =>
+              val currentBatchResult =
+                Chunk.fromIterable(
+                  respponse.getParameters.asScala.toList
```
afsalthaj, line 36: Typo in respponse

ithinkicancode: Fixed

afsalthaj: Probably we wanna move what we have as test to zio-config.
Although a containerised test in zio-config will hurt at the same time.
So its important you test this version against the test that you already have.

afsalthaj (approved):

Issue #951, opened 2022-11-21
Integrate with config front-end in ZIO 2.0.4


jdegoes:
ZIO 2.0.4 provides a config front-end, which can be used to describe the format of configuration data.

_ZIO Config_ should integrate cleanly into this configuration front-end, providing the ability to supply `ConfigProvider` for the front-end, as well as declaratively specify how multiple, disparate, disjointed or maybe even overlapping configuration sources collectively provide all of the configuration requirements of the application.

Several points of integration must be determined:

1. Should `zio.Config` replace `ConfigDescriptor`, or should it be possible to derive `zio.Config` from `ConfigDescriptor`?
2. If we delete `ConfigDescriptor`, how does this affect read, document, generateReport, and write?
3. Do we need `ConfigSource`, or can we instead use the `ConfigProvider`, maybe in a layer?
4. What does the user interface look like for using _ZIO Config_ providers?
5. Can _ZIO Config_ derive `Config` from data types, just like it can now derive `ConfigDescriptor` from ADTs?

vagroz:
It would be nice to have an interop with `pureconfig`'s `ConfigReader`.

jdegoes:
@afsalthaj and I discussed the following plan:

1. Remove everything in `zio.config._`, except for the documentation part, which can be retrofitted to work with `zio.Config` and `zio.ConfigProvider`.
2. Refactor integrations to use `Config` instead of `ConfigDescriptor`, and `ConfigProvider` instead of `ConfigSource`.
3. Possibly, add missing methods to `zio.ConfigProvider` or `zio.Config`.
4. Refactor each integration into a layer `ZLayer[Any, Nothing, Unit]`, which installs its config provider. So, for example, the hocon integration would expose a layer that can be incorporated into an app to install the HOCON config provider.
5. Release this as ZIO Config 4.0.0.

PR #1035, opened 2023-01-21
Work with new ZIO frontend and release 4.0.0

afsalthaj:
# Solves #951 

The `ConfigDescriptor` is moved to ZIO, now called as ZIO.Config.
The `ConfigSource` is now `ConfigProvider` in ZIO.

The PR integrates the library with these new changes in ZIO Core, and on a high level, aims to complete:

1. Backends such as HOCON to work with Config and ConfigProvider (and avoid the intermediate PropertyTree)
2. Auto-derivations

# Features  that could be removed
------------------------------------

_This comparison is mainly intended to extend ZIO core config (if needed)_

## Write back functionality

ZIO Config 3.x, for example could load up a HOCON file and write it back as system environment variables with customisable keys. This feature is removed.

----------------------

## No information about `ConfigSource` in documentation

ZIO Config 3.x used to provide Source information in documentation. This is because `ConfigSource` was part of the `ConfigDescriptor` language itself. With ZIO Config 4.x, we no more have details of `ConfigSource` (aka `ConfigProvider`) in documentation (unless we reimplement this feature) 

----------------------

## Less documentation details around keys in config sources

ZIO Config 3.x allows users to do things like the following example, which allowed more details in documentation, most importantly the structure of the keys based on source. With 4.x, though we have `Keys` available  as part of the `Config` language, its hard to know its format corresponding to specific sources

```scala

val rawConfig = 
  descriptor[A]
   
val config1 = 
   rawConfig.mapKey(_.toLowerCase) from  ConfigSource.fromJson(str)
   
val config2 = 
  rawConfig.mapKey(_.toUpperCase) from  ConfigSource.fromSystemEnv()
  
val config = config1.orElse(config2)

generateDocs(config) // lists all of the above details in markdown.   
```
 

 Now markdown exists, but we won't have these many details until we reimplement some of the features back again.

----------------------

## Lazy ConfigSource load

`configSource.memoize` in ZIO-Config 3.x ensures we have a single IO to load all values in a `product`. If not memoized, it will invoke separate IO for each field in the product. Non memoized version can be sometimes useful, if talking to ConfigSource is a heavy lifting IO operation.  

Example: If loading up a config value requires a DynamoDb scan , or querying database table, it can choose to load each field lazily (example: using separate connection if database). Although lazily loaded, we can define a `ConfigDescriptor` once and for all. `string("USERNAME").zip(string("PASSWORD")).from(dynamoDb)`. If we don't need separate initiation of AWS client, we could do  `string("USERNAME").zip(string("PASSWORD")).from(dynamoDb.memoize)`

This feature will not be present in the initial version of ZIO-Config 4.x since ConfigProvider hasn't had these embedded, unless piggybacking on ZIO.

----------------------

## StrictlyOnce ConfigSource load

`configSource.strictlyOnce` in ZIO-Config 3.x functionality ensures we are making a single IO to ConfigSource once and for all regardless of the number of times it is used in the application under ZIO.

Example: 

  ```scala
    val source: ConfigSource = ConfigSource.fromDatabase()
    
   val io = read(config from source)
   val io2 = read(anotherConfig from source)
  `io *> io2` // involves reading with the source only once  
  ```

This feature will not be present in the initial version of ZIO-Config 4.x.

----------------------

## `gen` module is removed. Auto generation of sample config isn't possible with ZIO-Config 4.x

In ZIO-Config 3.x, we had TransormOrFail node that was bidirectional And knows how to convert a type to its raw source format.  Reading this generated config back using “to” function hardly fails when used with DeriveGen of zio.test since we randomly generate types first  (example: Int) and write back and succeeds 99% of the times. This is how `gen` module worked.  Now we have `MapOrFail` in ZIO-Config 4.x  instead of `TransformOrFail`, and therefore even if we generate types, we hardly knows how to convert them to its raw format. Therefore gen module is removed.

We are thinking of keeping zio-config 3.x live for a bit more until we are so sure, no-one uses gen module.

----------------------

## Effectful ConfigSource

ConfigSource is now ConfigProvider, and it doesn't inherently assume effect ZIO now. Still thinking how to bring this back without hurting user friendliness.  

In 3.x,  the fact that loading a `ConfigSource` itself could fail was encoded as part of the `ConfigSource` language itself.

This was a feature later added to ZIO-Config to avoid users having to do `ZIO.attempt(loadHocon).mapError(ReadError.SourceError(_)).map(source => read(config from source)`.

With `ConfigProvider` / ZIO-Config 4.x, ConfigProvider by itself isn't effectful and the feature may be removed.

We may be switching back to recommending the users do `ZIO.attempt(ConfigProvider.fromHoconString(str)).map(error => Config.Error.SourceError(error)).map(_.load(config))`. 

----------------------

## Removed Property based round trip tests
Also removed a good amount of property based tests that worked based on write back functionality, where reading a source to a `ConfigSource` and writing it back using the same `ConfigDescriptor`gives the same Source data.

More to come...

Some API docs have also been removed.

## Progress

- [x]  Implement collectAll aka sequence (the core of auto derivation)

- [x] IndexedFlat will allow more natural implementation of ConfigProvider for complex Sources 

- [x] Implement Autoderivation basic for ZIO Config

- [x] Autoderivation with customisable nested path for ADTs - Scala3

- [x] Autoderivation with customisable nested path for ADTs - Scala2

- [x] Make a ConfigProvider out of a subsection of an existing ConfigProvider aka `at` functionality

- [x] Reimplement document functionality

- [x] mapKey functionality

- [x] update docs/website

- [x] Autoderivation for Scala3

- [ ] Layer integration

- [x] Autoderivation with support for type field in products (pureconfig style)

- [x] Implement prettyPrint for error message

- [x] Implement backend for typesafe config and update an example

- [ ] Update typesafe config tests

- [x] Implement backend for yaml

- [x] Update all examples and make sure they run successfully (10%)

- [x] Integration with cats, scalaz

- [x] Integration with refined

- [x] Make sure to release refined for scala 3.0.2

- [x] Integration with enumeratum

- [x] Reimplement enumerate in Config

- [x] ~~Try to integrate with gen module~~

- [x] Integration with aws-zio

- [ ] ~~absence of a value is considered as empty list instead of missing value~~ (not zio-config's concern anymore)

- [x] Remove the need to wrap the name of the sealed trait itself

- [ ] Make sure scala2 auto derivation and  scala 3 auto derivations are consistent in its behaviour. Currently it isn't!


core/shared/src/main/scala/zio/config/ReadModule.scala
```diff
@@ -7,13 +7,21 @@ import zio.{IO, Scope, ZIO}
 import scala.collection.mutable.{Map => MutableMap}
 
 import PropertyTreePath.Step
+import zio.ConfigProvider
+import zio.Config
+
+// Backward compatible approach to minimise the client changes
```
afsalthaj, line 13: Only for testing!

examples/shared/src/main/scala/zio/config/examples/magnolia/AutoDerivationSimple.scala
```diff
@@ -27,35 +28,35 @@ object MyConfig {
...
-        "aws.credentials.Token.value" -> "token",
-        "port"                        -> "10",
-        "default"                     -> "12",
-        "dburl.value"                 -> "some url",
-        "amount"                      -> "3.14",
-        "quanity"                     -> "30.0",
-        "price.Inr.value"             -> "1000",
-        "anotherDefault"              -> "14"
+        "aws.region"                              -> "us-east",
+        "aws.credentials.Credentials.Token.value" -> "token",
```
afsalthaj, line 42: Auto derivation currrently assumes no change from the client

PR #1066, opened 2023-02-23
Avoid scala-2.11 for all projects

afsalthaj:
Make sure there are no references to Scala211


afsalthaj:
Merging this to fix releases

PR #1091, opened 2023-02-28
Apply ZIO SBT Ecosystem Plugin

khajavi:


khajavi:
This work is ready for review.

The failing CI seems related to this issue #1095

afsalthaj:
Thanks Milad. Have been waiting for this plugin and great addition :-)

.github/workflows/ci.yml
```diff
@@ -1,89 +1,215 @@
-name: CI
+# This file was autogenerated using `zio-sbt-ci` plugin via `sbt generateGithubWorkflow` 
+# task and should be included in the git repository. Please do not edit it manually.
```
afsalthaj, line 2: 👍

.scalafmt.conf
```diff
@@ -1,16 +1,19 @@
...
+danglingParentheses.preset = true
 optIn.annotationNewlines = true
 
 rewrite.rules = [RedundantBraces]
 project.excludeFilters = [
   "core/src/main/scala/zio/config/ProductBuilder.scala",
-  "core/src/test/scala/zio/config/ProductBuilderTest.scala"
+  "core/src/test/scala/zio/config/ProductBuilderTest.scala",
+  "core/shared/src/main/scala-3.x/*",
+  "magnolia/shared/src/main/scala-dotty/*"
```
afsalthaj: I think this can also go into a custom module within ecosystem plugin that wraps scalafmt
And for custom settings, such as exclude, I could simply configure with in each module setting, `excludeFmt`

build.sbt
```diff
@@ -1,11 +1,13 @@
...
 
 inThisBuild(
   List(
-    organization := "dev.zio",
-    homepage := Some(url("https://zio.dev/zio-config/")),
-    licenses := List("Apache-2.0" -> url("http://www.apache.org/licenses/LICENSE-2.0")),
-    developers := List(
+    name                       := "ZIO Config",
+    crossScalaVersions -= scala211.value,
+    ciEnabledBranches          := Seq("series/4.x"),
```
afsalthaj: I think we should include `master` as well, since `master` works with latest zio

build.sbt
```diff
@@ -1,11 +1,13 @@
...
+crossScalaVersions := Seq.empty
 
 inThisBuild(
   List(
-    organization := "dev.zio",
-    homepage := Some(url("https://zio.dev/zio-config/")),
-    licenses := List("Apache-2.0" -> url("http://www.apache.org/licenses/LICENSE-2.0")),
-    developers := List(
+    name                       := "ZIO Config",
+    crossScalaVersions -= scala211.value,
```
afsalthaj, line 10: I would rather just rely on ecosystem plugin to decide the common scala versions that an ecosystem library should support. Obviously, ecosystem libraries can choose to exclude from this common list. This implies there will not be specific `inThisBuild` settings for crossScalaVersions. The exclusion, should be within each module.
Ecosystem Plugin can also choose to pre-built excluded list.

build.sbt
```diff
@@ -18,31 +20,30 @@ inThisBuild(
...
+      (zioConfigYamlJVM / thisProject).value.id                  -> (zioConfigYamlJVM / crossScalaVersions).value,
+      (zioConfigXmlJVM / thisProject).value.id                   -> (zioConfigXmlJVM / crossScalaVersions).value
+    ),
+    scalafixScalaBinaryVersion := ""
   )
 )
 
-addCommandAlias("fmt", "; scalafmtSbt; scalafmt; test:scalafmt")
-addCommandAlias("fix", "; all compile:scalafix test:scalafix; all scalafmtSbt scalafmtAll")
 addCommandAlias("compileAll", "; ++2.12.16; root2-12/compile; ++2.13.8!; root2-13/compile")
```
afsalthaj: I would delete this as well. CommandAlias can be a pain. I would either choose it as sbtSettings  that are not based on String. Even better is within ecosystem plugin

build.sbt
```diff
@@ -18,31 +20,30 @@ inThisBuild(
...
 addCommandAlias("testAll", "; ++2.12.16; root2-12/test; ++2.13.8!; root2-13/test")
```
afsalthaj: I would rather choose the same Same as above

build.sbt
```diff
@@ -65,13 +67,14 @@ lazy val magnoliaDependencies =
 
 lazy val refinedDependencies =
   libraryDependencies ++= {
-    if (scalaBinaryVersion.value == "2.11") Seq.empty // Just to make IntelliJ happy
+    if (scalaVersion.value == scala211.value) Seq.empty // Just to make IntelliJ happy
```
afsalthaj: This can happen several times. One setting can be `addDependency` which takes simply takes a list of scala versions to be excluded. Otherwise, different projects may have different ways to handle dependency settings. 

Also `refinedDependencies` is an unnecessary redirection, that it could be inlined as well

build.sbt
```diff
@@ -138,116 +141,91 @@ lazy val `root3` =
...
-    crossScalaVersions --= Seq("2.11"),
     libraryDependencies ++= Seq(
-      "com.amazonaws" % "aws-java-sdk-ssm" % awsVersion,
-      "dev.zio"      %% "zio-streams"      % zioVersion,
-      "dev.zio"      %% "zio-test"         % zioVersion % Test,
-      "dev.zio"      %% "zio-test-sbt"     % zioVersion % Test
-    ),
-    testFrameworks := Seq(new TestFramework("zio.test.sbt.ZTestFramework"))
+      "com.amazonaws" % "aws-java-sdk-ssm" % awsVersion
+    )
```
afsalthaj, line 169: Much better for any new contributor, is to be able to say

```scala
lazy val zioConfigAws = jvmModule("aws").withDependencies().dependOn(zioConfig)
```

As mentioned above withDependencies can take care of weird cases, such as excluding if Scala-version is `x` etc.

The ecosystem plugin mandates the name of the root project (say zio-config), and any module definition simply append the file path name. This way every project becomes a standard

build.sbt
```diff
@@ -272,176 +250,139 @@ lazy val examplesJVM = examples.jvm
...
-  .settings(dottySettings)
+lazy val zioConfigXmlJVM = zioConfigXml.jvm.settings(scala3Settings)
 
-lazy val zioConfigScalaz    = crossProject(JSPlatform, JVMPlatform, NativePlatform)
+lazy val zioConfigScalaz = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .in(file("scalaz"))
-  .settings(stdSettings("zio-config-scalaz"))
-  .settings(crossProjectSettings)
+  .settings(stdSettings(name = "zio-config-scalaz", enableCrossProject = true))
+  .settings(enableZIO())
```
afsalthaj, line 300: I think its mostly to help defining modules easily without relying on adhoc sbt settings. As mentioned above, as a maintainer I am far relieved if I can just simply define `module(...)` using the plugin, or if needed `jvmModule`, `jsModule` etc.

afsalthaj:
I have put some comments, but all of it boils down to 1 suggestion, that being able to define module based on the plugin, 

```scala
lazy val myModule = 
   module("aws")
    .withDependencies("dev.zio" %%% "izumi-reflect" % "2.2.2")
    .dependsOn(core)
```

Adding customisations becomes even simpler

```scala
lazy val myModule = 
   module("aws")
    .skipScala211
    .withDependencies("dev.zio" %%% "izumi-reflect" % "2.2.2")
    .dependsOn(core)

```

This is much easier for a maintainer. Example: In the above case, `core` can support 211, but `aws` cannot, and there is no need of an indirection such as having a big Map of each module and scala versions supported

.scalafmt.conf
khajavi: Thank you! Added to the backlog.

build.sbt
khajavi: What about series/2.x? Doesn't it work for the latest version of ZIO?

khajavi:
@afsalthaj 
That is an excellent DSL for writing builds without any boilerplate. I agree with such language. The first iteration aimed to factor out `BuildHelper` and make it maintainable. 
I am excited to work on implementing your suggestions for the next phase.

khajavi:
> In the above case, `core` can support 211, but `aws` cannot, and there is no need of an indirection such as having a big Map of each module and scala versions supported

The name of `supportedScalaVersions` is somehow confusing, I will rename it to something like `ciScalaVersionMatrix`. We just use that to create the test matrix. This is just a **workaround** because I lack deep knowledge of how to iterate through sbt submodules and create matrix map pragmatically.

So ideally I would like to have something like this:

```scala
val ciScalaVersionsMatrix: Map[String, Seq[String]] = 
  listOfAllModules.map { module =>
      (module / thisProject).value.id   -> (module / crossScalaVersions).value
  }
```

But it will not compile due to complicated macro-based sbt internals. We need to find sbt specific solution to traverse modules and pull out settings.

build.sbt
afsalthaj: I don't think we need to support 2.x going forward

build.sbt
khajavi: Currently, zio-sbt provides default scala versions, then we can add/remove and customize that for the whole modules or any submodule. 
I'm not sure I got what you said, here I excluded scala 211 from all modules, although we can do that one by one.

build.sbt
khajavi: That's a good idea, but it is the first time I saw such a case. I'm doing the same for other ecosystem projects. If I saw that the common case, I'll make a general utility inside zio-sbt.

build.sbt
khajavi: I agree, will work on this!

build.sbt
khajavi: Done!

build.sbt
khajavi: Done!

build.sbt
khajavi: Done!

khajavi:
Thrilled to see the geeeen color 😅
@afsalthaj Can you please have a look?

build.sbt
khajavi: Just wrote these utility functions:

```scala
def addDependenciesFor(scalaBinaryVersions: String*)(dependencies: ModuleID*) =
  libraryDependencies ++= {
    if (scalaBinaryVersions.contains(scalaBinaryVersion.value)) dependencies else Seq.empty
  }

def addDependenciesExceptFor(scalaBinaryVersions: String*)(dependencies: ModuleID*) =
  libraryDependencies ++= {
    if (scalaBinaryVersions.contains(scalaBinaryVersion.value)) Seq.empty else dependencies
  }

def addScalacOptionsFor(scalaBinaryVersions: String*)(options: String*) =
  scalacOptions ++= {
    if (scalaBinaryVersions.contains(scalaBinaryVersion.value)) options else Seq.empty
  }

def addScalacOptionsExceptFor(scalaBinaryVersions: String*)(options: String*) =
  scalacOptions ++= {
    if (scalaBinaryVersions.contains(scalaBinaryVersion.value)) Seq.empty else options
  }
```

afsalthaj (dismissed):

PR #1114, opened 2023-03-09
Disable comma separation in values in ZIO for complex configs handled by zio-config

afsalthaj:
ZIO-Config currently depends fully on ZIO library for its indexed config formats too.
However, as a consequence we reintroduced certain bugs related to the fact that values were considered as a list if there exists a comma. This can hurt a lot of users, who would normally consider list as something that hinted by specific characters such a starting `[` and close `]` in Yaml or `-` in YML  . This implies, for example, users expects only `["a", "b", "c"]` to be a list and not `"a, b, c"`. If they need `"a,b,c"` to be considered as a list, then they can set `enableCommaSeparatedValueAsList = true`. The name is long to understand what's going on.

With ZIO-3.0, we can avoid this slight weirdness


typesafe/shared/src/main/scala/zio/config/typesafe/TypesafeConfigProvider.scala
```diff
@@ -59,7 +58,8 @@ object TypesafeConfigProvider {
 
     ConfigProvider.fromMap(
       indexedMapWithHiddenDelimiter,
-      pathDelim = hiddenDelim
+      pathDelim = hiddenDelim,
+      seqDelim = if(enableCommaSeparatedValueAsList) "," else hiddenDelim
```
guersam, line 62: Would it be better if the user can provide their own separator like `|` ? For example, with `valueSeparator: Option[String]` this line can be like this:

```scala
seqDelim = valueSeparator.getOrElse(hiddenDelim)
```

PR #1441, opened 2024-07-19
chore: bump scala versions (2.12.19, 2.13.13, 3.3.3 LTS)

vladimir-lu:


vladimir-lu:
For whatever reason, some of the examples seem to be failing, e.g.: https://github.com/zio/zio-config/blob/209e0ac098f6849763ae3ccfc3f2fac4fde86bcd/examples/shared/src/main/scala/zio/config/examples/autoderivation/AutoDerivationPureConfig.scala#L81

I don't get why it should work - the discriminator is not optional or is it? 

Edit: never mind, that example doesn't work on the current HEAD of the origin/4.x branch either. Shall I fix it in this PR?

kyri-petrou:
Closing as this was addressed in https://github.com/zio/zio-config/pull/1478

PR #1455, opened 2024-09-05
add DeriveConfig instances for NonEmptyChunk

mberndt123:


mberndt123:
CI is failing because I need a newer ZIO version and that seems to be incompatible with Java 8. Maybe it's time to throw it out?

vigoo (approved):

vigoo:
> CI is failing because I need a newer ZIO version and that seems to be incompatible with Java 8. Maybe it's time to throw it out?

Yes, please change the CI to test with the same set of Java versions as core ZIO does

mberndt123:
@vigoo 

It's passing CI now, so I think it's ready to merge

PR #1464, opened 2024-10-03
Update SBT for pureconfig.

yangzai:
It seems like the `zio-config-pureconfig` module never got published.

I'm not sure if the issue is coming from the SBT config.

In any case this updates the SBT config, in particular it switches the dependency from `pureconfig` to `pureconfig-core` so it should now be able to support Scala 3.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -34,15 +34,15 @@ addCommandAlias(
 )
 addCommandAlias(
   "testJVM212",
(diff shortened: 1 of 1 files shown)


yangzai:
@987Nabil should I backport this to other branches that are still in maintenance?

987Nabil:
No. I don't expect anything else then 4.x will be released

PR #1466, opened 2024-10-06
chore/refactor to projectmatrix

ThijsBroersen:
PR to update about all...
Build has been migrated to sbt-projectmatrix because the existing build had quite some issues and did target all platforms and versions correctly. 

I also noticed some gaps. Like annotations not implemented for Scala 3. 

The JDK version is updated to 17 setting the jvm release target to 11.

Fixes/enables scalafix on CI (which is enabled by default with zio-sbt-ci)


ThijsBroersen:
I reverted the zio-sbt-website version. As the new version added a `+publishLocal` step which causes an issue when using projectmatrix. It needs to be without the `+`, I did not know how to override it.
For now it targets only Scala 2.13. I think this is also what main did as 2.13 was the default scalaVersion.

examples/src/main/scala/zio/config/examples/autoderivation/AutoDerivationPureConfig.scala
```diff
@@ -78,19 +78,19 @@ object AutoDerivationSealedTraitPureConfig extends App with EitherImpureOps {
...
-    """
-      |field = randomcustomname
-      |""".stripMargin
+  // val s3: String =
+  //   """
+  //     |field = randomcustomname
+  //     |""".stripMargin
 
-  assert(read(deriveConfig[AwsConfig] from TypesafeConfigProvider.fromHoconString(s3)) equalM AwsConfig(SubObject2))
+  // assert(read(deriveConfig[AwsConfig] from TypesafeConfigProvider.fromHoconString(s3)) equalM AwsConfig(SubObject2))
```
ThijsBroersen: Still need to check these.

examples/src/main/scala/zio/config/examples/autoderivation/AutoDerivationSimple.scala
```diff
@@ -42,7 +42,7 @@ object AutomaticConfig extends App {
         "aws.credentials.Token.value" -> "token",
         "port"                        -> "10",
         "default"                     -> "12",
-        "dburl"                       -> "some url",
+        "dburl.value"                 -> "some url",
```
ThijsBroersen, line 45: There is some inconsistency on the usage of value-classes. Sometimes AnyVal is extended, sometimes not. In some tests it expects the path to have the nested path, but in this code it did not. 
I think also some apps were not working in in the main branch. This change for now makes it work.

examples/src/main/scala/zio/config/examples/autoderivation/AutoDerivationPureConfig.scala
ThijsBroersen: re-enabled, in main branch series/4.x it was also failing

project/plugins.sbt
```diff
@@ -1,22 +1,16 @@
...
+addSbtPlugin("com.typesafe"                      % "sbt-mima-plugin"           % "1.1.4")
+addSbtPlugin("de.heikoseeberger"                 % "sbt-header"                % "5.10.0")
+addSbtPlugin("org.scala-js"                      % "sbt-scalajs"               % "1.17.0")
+addSbtPlugin("org.scala-native"                  % "sbt-scala-native"          % "0.5.5")
+addSbtPlugin("org.scalameta"                     % "sbt-mdoc"                  % "2.6.1")
+addSbtPlugin("org.scalameta"                     % "sbt-scalafmt"              % "2.5.2")
+addSbtPlugin("pl.project13.scala"                % "sbt-jcstress"              % "0.2.0")
+addSbtPlugin("pl.project13.scala"                % "sbt-jmh"                   % "0.4.7")
+addSbtPlugin("dev.zio"                           % "zio-sbt-website"           % "0.3.1")
+addSbtPlugin("com.eed3si9n"                      % "sbt-projectmatrix"         % "0.10.0")
```
987Nabil: I am not sure about this. I don't think any other zio project is using this so far. And we have zio-sbt to help us with the builds. I'd prefer to stick to the same solution, unless you can make a compelling argument.

project/plugins.sbt
```diff
@@ -1,22 +1,16 @@
...
-addSbtPlugin("org.portable-scala"                % "sbt-scalajs-crossproject"      % "1.2.0")
-addSbtPlugin("org.scala-js"                      % "sbt-scalajs"                   % "1.16.0")
-addSbtPlugin("org.scala-native"                  % "sbt-scala-native"              % "0.4.16")
-addSbtPlugin("org.scalameta"                     % "sbt-mdoc"                      % "2.3.7")
-addSbtPlugin("org.scalameta"                     % "sbt-scalafmt"                  % "2.4.6")
-addSbtPlugin("pl.project13.scala"                % "sbt-jcstress"                  % "0.2.0")
-addSbtPlugin("pl.project13.scala"                % "sbt-jmh"                       % "0.4.4")
-addSbtPlugin("dev.zio"                           % "zio-sbt-website"               % "0.3.1")
-
-libraryDependencies += "org.snakeyaml" % "snakeyaml-engine" % "2.6"
```
987Nabil: was unused? or is now a transitive dependency?

project/plugins.sbt
ThijsBroersen: This sbt-projectmatrix makes the build much nicer to work with imo. It is more explicit and helps to discover what platforms and versions are in scope. I have seen multiple public projects already using it and I am using it in all of my customer projects. It make it much easier to crossbuild, export all in a single bloopInstall instead of just one Scala version.

ThijsBroersen: To my understanding it should also be a default feature in SBT 2.x

project/plugins.sbt
ThijsBroersen: It was used to extract Scale versions from the `ci.yml` definition. There is quite some difference on how different zio-projects are being configured. The `BuildHelper.scala` has a lot if different implementations. I think the Scala versions should just be part of the build files and not extracted from ci.yml. Most projects do this and it prevent us from requiring this dependency in the sbt tasks.

project/plugins.sbt
ThijsBroersen: With the existing build and the command aliases to target the different platforms and scala versions it was quite error prone to make sure all projects were included. Now you do not have to use the `+` or `++` anymore. All projects have a direct projectref in the build.

ThijsBroersen: I think the build is much more concise now and it is targeting all the platforms and Scala versions it can support (the existing build was missing some).

ThijsBroersen: ref sbt-projectmatrix insource in sbt 2.x: https://eed3si9n.com/sbt-2.0.0-beta

PR #1471, opened 2024-10-16
Feat/support all annotations in scala 3

ThijsBroersen:
This PR add support for annotations in Scala 3 on all platforms. The added tests are used for testing all platforms and scala versions to ensure unambiguous behaviour. 
This PR introduces the `suffix` annotation as a replacement for `postfix` (to deprecate). Semantics of `postfix` to not properly reflect the feature.


ThijsBroersen:
@kyri-petrou @afsalthaj 
I see mima checks got recently added. I wanted to share KeyModifier between Scala 2 and 3 but somehow that broke mima. I now duplicated the code.

How important is mima atm? I changed Macros to AnnotationMacros and made it private to the magnolia package. I don't think it should be public api.
Should the CI Mima step hard fail atm, or warn? Or perhaps a manual approval? (e.g. https://github.com/marketplace/actions/manual-workflow-approval)?

magnolia/shared/src/main/scala-dotty/zio/config/magnolia/DeriveConfig.scala
```diff
@@ -273,23 +309,37 @@ object DeriveConfig {
...
+    val modifyKey: String => String =
+      sortedKeyModifiers.map(KeyModifier.getModifierFunction).foldLeft(_)((key, modifier) => modifier(key))
+
     alternativeKeys match {
-      case Nil  => originalKey.fold(desc)(desc.nested(_))
-      case keys => keys.view.map(desc.nested(_)).reduce(_ orElse _)
+      case Nil  =>
+        originalKey.fold(desc)(k => desc.nested(modifyKey(k)))
+      // case keys => keys.view.map(k => desc.nested(modifyKey(k))).reduce(_ orElse _) // Looks like the Scala 3 implementation modifies alternative names while the Scala 2 implementations treats them as is.
+      case keys => keys.view.map(k => desc.nested(k)).reduce(_ orElse _)
```
ThijsBroersen, line 342: please review this carefully, because I think L341 should be enabled and L342 disabled. This changes the existing behaviour only to match the Scala 2 behaviour.

ThijsBroersen: L340 I already adjusted.

PR #1479, opened 2024-11-06
Make it possible to use `derives Config` syntax in Scala 3

kyri-petrou:
See related issue in ZIO: https://github.com/zio/zio/issues/9268.

This PR allows essentially allows users to use `derives Config` syntax on Product and Sum types when the following import is present:

```scala
import zio.config.magnolia.*
```


ghostdogpr (approved):
Very nice!

PR #1509, opened 2024-12-29
Fix compiling after merge error for Scala 3

kyri-petrou:
Seems that this was caused by a merge error. Not sure how did it made it past merge conflict detection.

Also fixing the Website CI since it seems broken at the moment


kyri-petrou:
Going to merge in to fix CI

